PR #2, opened 2022-02-09
Attempting to fix breakpoint construction from alignment blocks.

tfenne:
@nh13 This isn't ready to merge, I need to add the reverse strand version of several of the tests, and probably test some more things too.  But I think this is working.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
@@ -202,7 +202,7 @@ object SvPileup extends LazyLogging {
     if (r1.refIndex == r2.refIndex) None
     else {
       val evidence = if (seg1.origin.isPairedWith(seg2.origin)) ReadPairInterContig else SplitReadInterContig
-      Some(PutativeBreakpoint(seg1=seg1, seg2=seg2, evidence=evidence))
+      Some(PutativeBreakpoint(from=seg1, into=seg2, evidence=evidence))
     }
   }
 
@@ -222,7 +222,7 @@ object SvPileup extends LazyLogging {
     }
     if (innerDistance <= maxInnerDistance) None else {
       val evidence = if (seg1.origin.isPairedWith(seg2.origin)) ReadPairIntraContig else SplitReadIntraContig
-      Some(PutativeBreakpoint(seg1=seg1, seg2=seg2, evidence=evidence))
+      Some(PutativeBreakpoint(from=seg1, into=seg2, evidence=evidence))
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/com/fulcrumgenomics/fgsv/util/PutativeBreakpoint.scala
```diff
@@ -5,52 +5,69 @@ package com.fulcrumgenomics.fgsv.util
...
+      leftRefIndex  = b1.range.refIndex,
+      leftPos       = if (b1.positiveStrand) b1.range.end else b1.range.start,
+      leftPositive  = b1.positiveStrand,
+      rightRefIndex = b2.range.refIndex,
+      rightPos      = if (b2.positiveStrand) b2.range.start else b2.range.end,
+      rightPositive = b2.positiveStrand,
       evidence      = evidence
     )
+
+    if (b1.range < b2.range) x else x.copy(
```
tfenne, line 64: This isn't the most efficient way to do this, but it's the easiest to think about so for now I'm doing it.  Once I have enough tests to be confident it's all right, we can refactor to not create two breakpoints.

src/test/scala/com/fulcrumgenomics/fgsv/util/PutativeBreakpointTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    (bp.leftPos, bp.leftPositive)   shouldBe (149, true)
+    (bp.rightPos, bp.rightPositive) shouldBe (1000, true)
+
+    // And again as if we read it with a read on the negative strand
+    val revBlock1 = block(readStart=1,  refIndex=1, refStart=1000, length=30, positive=false)
+    val revBlock2 = block(readStart=31, refIndex=1, refStart=100,  length=50, positive=false)
+    val rbp = PutativeBreakpoint(revBlock1, revBlock2, EvidenceType.SplitReadIntraContig)
+
+    (rbp.leftPos, rbp.leftPositive)   shouldBe (149, true)
+    (rbp.rightPos, rbp.rightPositive) shouldBe (1000, true)
```
tfenne, line 33: Getting this right really hurt.

nh13: Real MVP

src/main/scala/com/fulcrumgenomics/fgsv/util/PutativeBreakpoint.scala
```diff
@@ -5,52 +5,81 @@ package com.fulcrumgenomics.fgsv.util
...
+      leftPos       = rightPos,
+      leftPositive  = !rightPositive,
+      rightRefIndex = leftRefIndex,
+      rightPos      = leftPos,
+      rightPositive = !leftPositive
+    )
+
+  /** Returns true if the representation of the breakend is canonical, with the left hand side of the break
+   * earlier on the genome than the right hand side. */
+  def isCanonical: Boolean = (leftRefIndex < rightRefIndex) ||
```
tfenne, line 61: I think we should think long and hard about what "canonical" should be, probably review the VCF spec and what callers do.  See the test for the tandem duplication as an example of where we might want to do things differently.

src/test/scala/com/fulcrumgenomics/fgsv/util/PutativeBreakpointTest.scala
```diff
@@ -0,0 +1,98 @@
...
+    // We want the same representation regardless of which piece of the duplication was sequenced more so
+    // we test where the second block is shorter, the same length as, and longer than the first
+    Seq(-1, 0, 1).foreach { addend =>
+      val block1Length = 50
+      val block2Length = block1Length + addend
+      val block1 = block(readStart=1,  refIndex=1, refStart=100, length=block1Length, positive=true)
+      val block2 = block(readStart=51, refIndex=1, refStart=100, length=block2Length, positive=true)
+      val bp     = PutativeBreakpoint(block1, block2, EvidenceType.SplitReadIntraContig)
+      (bp.leftPos,  bp.leftPositive)  shouldBe (100, false)
+      (bp.rightPos, bp.rightPositive) shouldBe (149, false)
```
tfenne, line 94: So the test case here is a tandem duplication of 50bp at positions 100-149.  So we basically read that chunk of genome twice sequentially, and make a breakend out of it.

Personally I would find it more intuitive to have the breakend information laid out so that it says:
- You're going along chr1 and get to base 149, and then
- Suddenly you find yourself back at base 100 again, still reading in the same direction

I.e. the left hand side would be `149+` and the right hand side would be `100+`, but given our existing definition of canonical as "left hand side lower" we use up with the left hand side being `100-` and the right hand being `149-`.  I _think_ if I had my druthers I would canonicalize so that the left hand side was always positive even if that meant having it be higher on the genome, but I haven't thought about it hard enough to know if that's a bad idea or not.

src/main/scala/com/fulcrumgenomics/fgsv/util/PutativeBreakpoint.scala
```diff
@@ -5,52 +5,81 @@ package com.fulcrumgenomics.fgsv.util
...
+    if (result == 0) result = this.rightPositive.compare(that.rightPositive)
     if (result == 0) result = EvidenceType.indexOf(this.evidence) - EvidenceType.indexOf(that.evidence)
     result
   }
+
+  /**
+   * Returns an inverted copy of the breakpoint where the left and right sides of the breakpoint have been
+   * switched.
+   */
+  def inverted: PutativeBreakpoint = copy(
```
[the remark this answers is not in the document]
tfenne: How about `reversed`?  I think that makes the most sense as all we're really saying is to reverse observation of it?

PR #3, opened 2022-02-11
Lots of cleanup of the initial bolus of code.

tfenne:
Lots of changes here, but I think it's all stuff I suggested addressing in reviewing #1 

@nh13 It took me a while to fixup #2 after rebasing on top of your changes to #1.  Can you please review this and let me get it merged before making any more changes to #1?  Thanks!


src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -43,114 +43,128 @@ class SvPileup
...
-    /** Returns the identifier of the breakpoint, if it exists, or the next identifier that should be used. */
-    def apply(breakpoint: PutativeBreakpoint): Long = {
-      // NB: use id=0 and evidence=any so we can find it in the map regardless of existing id and evidence type
-      this.breakpointToId.getOrElseUpdate(breakpoint.copy(id=0, evidence=EvidenceType.ReadPairTandem), {
-        val id = this._nextId
-        this._nextId += 1
-        id
-      })
+  /** Class that tracks IDs and counts for Breakpoints during discovery. */
+  private class BreakpointTracker {
```
nh13, line 136: I wonder if this is the simplest interface, given method below like `breakpoints`, `countsFor` and `evidenceFor` are present.  You could also implement `Map[Breakpoint, ...]` but you have two maps.  So perhaps it's fine the way it is.  Something to think about.
```suggestion
  private class BreakpointTracker extends Iterator[(Breakpoint, SimpleCounter[EvidenceType]] {
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -178,15 +191,11 @@ object SvPileup extends LazyLogging {
...
       segments.iterator.sliding(2).flatMap { case Seq(seg1, seg2) =>
-        var result: Option[PutativeBreakpoint] = determineInterContigBreakpiont(seg1=seg1, seg2=seg2)
-        if (result.isEmpty) {
-          val maxInnerDistance = if (seg1.origin != seg2.origin) maxAlignedSegmentInnerDistance else maxReadPairInnerDistance
-          result = determineIntraContigBreakpiont(seg1=seg1, seg2=seg2, maxInnerDistance=maxInnerDistance)
-        }
-        if (result.isEmpty) result = determineOddPairOrientation(seg1=seg1, seg2=seg2)
+        var result                 = determineInterContigBreakpoint(seg1, seg2)
+        if (result.isEmpty) result = determineIntraContigBreakpoint(seg1, seg2, maxAlignedSegmentInnerDistance , maxReadPairInnerDistance)
+        if (result.isEmpty) result = determineOddPairOrientation(seg1, seg2)
```
nh13, line 196: Shall we collect all evidence for this template, or just the first one?  I have it just the first one to make the summary statistics for each breakpoint have `sum(# of templates) == sum(# of evidences)`.  For annotating reads, I think I want to know all types of evidence that template implies though.

nh13 (approved):
Nothing controversial in my review.  Merge when you're happy with it.

I do think at some point we should annotate reads with all  possible evidence types that the read/template implies.  We could have the one used for summary stats be listed first.  So if you agree, we can add that as an issue for a later time.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
tfenne: It's possible there's a better formulation.  But for now the interface is pretty narrow, and it's private, so it'll be easy to change.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -43,114 +43,128 @@ class SvPileup
...
+      ns(EvidenceType.indexOf(ev)) += 1
+      id
     }
 
-    def nextId: Long = this._nextId
+    /** Returns an iterator over the set of observed breakpoints, ordering is not predictable. */
+    def breakpoints: Iterator[Breakpoint] = breakpointToId.keysIterator
+
+    /** Gets the ID of a breakpoint that has previously been tracked. */
+    def idOf(bp: Breakpoint): BreakpointId = this.breakpointToId(bp)
```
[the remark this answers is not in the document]
tfenne: Exactly - with it being private I'm sure I'm only calling it on the breakpoints I got from calling `.breakpoints()`

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
tfenne: I think given our discussion, we're in a good place.  We'll reduce evidence types down to `SplitRead` and `ReadPair` and that'll be it.

PR #4, opened 2022-02-15
Reducing EvidenceType to [SplitRead, ReadPair], adding lots of tests, support for unpaired data

tfenne:
This PR got a it bigger than I had originally intended.  There are a few major changes rolled into one PR:

1. Reducing `EvidenceType` down to just `SplitRead` and `ReadPair`, and everything that follows from that.  As part of this I restructured the `determine*` methods in SvPileup to be called `is*` and return booleans, as the construction of the `BreakpointEvidence` is the same in all cases and can be pulled up into the calling method.  I also collapsed `isIntraContig` and `isOddPair` because a) they've gotten simpler and b) I kept confusing myself as to which method should identify which kind of breakpoint where the ends are on the same contig (e.g. where does detection of the following belong: seg1=chr1:500:F, seg2=chr1:200:F`?)
2. While doing this I noticed what I think is a large inconsistency in how `AlignedSegment` extraction was working, and a related bug in how "odd pair" detection was working.  The short version is that depending on the branch you took, based on the number of supplementary alignments, R2 segments were sometimes strand flipped and sometimes not, making accurate odd-pair detection impossible.  And the odd-pair detection that was there relied on the segments from R2 _not_ being strand flipped.  I've cleaned this up so that R2 records are always strand flipped (except, in 4 below)
3. This made me really nervous so I wrote a whole bunch of tests that start at `findBreakpoints(Template)` so I could be sure that the whole process of converting `SamRecord`s to `AlignedSegment`s through to detection of breakpoints was working, and fixed up a few things I found along the way.  I'm not much more confident it's doing the right thing.
4. I would really like the tool to be able to make use of unpaired data and PE data where one end is either unmapped or has poor mapping quality.  E.g. if you have a large insertion next to a rearrangement causing say R2 to be unmapped, it would be good to be able to call the rearrangement from R1.  To this end I've pulled all read filtering up out of `AlignedSegment` and into `SvPileup.filterTemplate()` which checks to see which primary reads are acceptable and then filters those and the corresponding supplementary reads appropriately.  And added a bunch of tests for this.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
@@ -176,14 +176,14 @@ object AlignedSegment extends LazyLogging {
       case (None     , _    ) => NoSegments
     }
 
-    if (r1Segments.isEmpty)
-      r2Segments
-    else if (r2Segments.isEmpty)
-      r1Segments
-    else
-      // Assume the library should be FR, reverse the R2 segments and flip their strand so that they're
-      // in the same order and orientation as R1 segments
-      mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments.reverse.map(b => b.copy(positiveStrand = !b.positiveStrand)))
+    (r1Segments, r2Segments) match {
+      case (segs, IndexedSeq()) => segs
+      case (IndexedSeq(), segs) => segs
+      case (s1, s2) =>
+        // Assume the library should be FR, reverse the R2 segments and flip their strand so that they're
+        // in the same order and orientation as R1 segments
+        mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments.reverse.map(b => b.copy(positiveStrand = !b.positiveStrand)))
+    }
   }
 
   /**
@@ -240,7 +240,7 @@ sealed trait SegmentOrigin extends EnumEntry {
   }
 }
 
-/** Enumeration for where a given [[AlignedSegment]] origintes. */
+/** Enumeration for where a given [[AlignedSegment]] originates. */
 object SegmentOrigin extends FgBioEnum[SegmentOrigin] {
   def values: IndexedSeq[SegmentOrigin] = findValues
   /** R1-only segment */
--- a/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
@@ -54,8 +54,6 @@ case class AlignedSegment(origin: SegmentOrigin,
 object AlignedSegment extends LazyLogging {
   private val NoSegments: IndexedSeq[AlignedSegment] = IndexedSeq.empty
 
-  private def fail(message: String) = throw new IllegalStateException(message)
-
   /** Builds an alignment segment from a [[SamRecord]]
    *
    * @param rec the mapped record
@@ -92,7 +90,7 @@ object AlignedSegment extends LazyLogging {
    *                            set of segments.
    */
   def segmentsFrom(primary: SamRecord,
-                   supplementals: Iterator[SamRecord],
+                   supplementals: Seq[SamRecord],
                    minUniqueBasesToAdd: Int): IndexedSeq[AlignedSegment] = {
     require(supplementals.nonEmpty)
     val primarySegment = AlignedSegment(primary)
@@ -115,12 +113,12 @@ object AlignedSegment extends LazyLogging {
    *                            set of segments.
    */
   def segmentsFrom(primary: AlignedSegment,
-                   supplementals: Iterator[AlignedSegment],
+                   supplementals: Seq[AlignedSegment],
                    readLength: Int,
                    minUniqueBasesToAdd: Int): IndexedSeq[AlignedSegment] = {
     require(supplementals.nonEmpty)
-    val supplSegments = supplementals.toIndexedSeq.sortBy(b => (b.readStart, b.readEnd))
-    val builder     = IndexedSeq.newBuilder[AlignedSegment]
+    val supplSegments = supplementals.sortBy(b => (b.readStart, b.readEnd))
+    val builder       = IndexedSeq.newBuilder[AlignedSegment]
 
     require(supplementals.forall(_.origin == primary.origin))
 
@@ -148,9 +146,9 @@ object AlignedSegment extends LazyLogging {
 
   /** Builds alignment segments from a template.
    *
-   * If no supplementary alignments exist, then an index segment for R1 and R1 respectively is returned.  Otherwise,
-   * supplementary alignments are filtered to have a minimum mapping quality, segments built for reach, and then the
-   * following iterative algorithm is performed for each read end:
+   * If no supplementary alignments exist, then a segment for each primary record (R1 and/or R2) is returned. Otherwise,
+   * segments are built for each supplementary alignment, and then the following iterative algorithm is performed for
+   * each read end:
    * 1. The primary alignment for the given read end is added to the set to keep.
    * 2. Next, the alignment segment for the supplementary alignment that maps the left-most base in sequencing order is
    * chosen.  If it maps at least the `minUniqueBasesToAdd` new bases versus those alignment segments already picked,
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -147,43 +146,44 @@ object AlignedSegment extends LazyLogging {
...
-  def segmentsFrom(template: Template,
-                   minSupplementaryMappingQuality: Int,
-                   minUniqueBasesToAdd: Int): IndexedSeq[AlignedSegment] = {
-    val r1 = template.r1.getOrElse(fail(s"No R1 for template: ${template.name}"))
-    val r2 = template.r2.getOrElse(fail(s"No R2 for template: ${template.name}"))
-    val r1Supps = template.r1Supplementals.iterator.filter(_.mapq >= minSupplementaryMappingQuality)
-    val r2Supps = template.r2Supplementals.iterator.filter(_.mapq >= minSupplementaryMappingQuality)
+  def segmentsFrom(template: Template, minUniqueBasesToAdd: Int): IndexedSeq[AlignedSegment] = {
+    val r1Primary = template.r1.filter(_.mapped)
+    val r2Primary = template.r2.filter(_.mapped)
```
tfenne, line 164: The filtering here is redundant, but I think makes the intent clearer.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -95,8 +94,8 @@ class SvPileup
   /** Coalesce the breakpoint counts and write them. */
   private def writeBreakpoints(tracker: BreakpointTracker, dict: SequenceDictionary): Unit = {
     val writer = Io.toWriter(output)
-    val fields = Seq("id", "left_contig", "left_pos", "left_strand", "right_contig", "right_pos", "right_strand", "total") ++
-      EvidenceType.values.map(_.snakeName)
+    val fields = Seq("id", "left_contig", "left_pos", "left_strand", "right_contig", "right_pos", "right_strand",
+      "split_reads", "read_pairs", "total")
```
tfenne, line 98: Now that we have a fixed (2) set of evidence types, we could convert this to a `Metric` which would make reading it back and working on it later much easier.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
-    /** Gets the evidence counts for a given breakpoint zipped with the evidence types. */
-    def evidenceFor(bp: Breakpoint): IndexedSeq[(EvidenceType, Int)] = {
-      EvidenceType.values.zip(this.counts(bp))
-    }
+  /**
+   * Performs filtering on a Template to remove primary records that are unmapped and then remove
+   * supplementary records if there is no matching primary record retained.  If neither primary
+   * record is retained, returns None, else returns Some(Template).
+   */
+  def filterTemplate(t: Template,
```
tfenne, line 174: This could definitely be optimized a bit in order to not create copies of things when not needed, but that would make it more complicated and I think this is fine for now.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
-    )
+    val segments = AlignedSegment.segmentsFrom(template, minUniqueBasesToAdd = minUniqueBasesToAdd)
+
+    segments.length match {
+      case 0 | 1 =>
+        NoBreakpoints
+      case 2     =>
+        // Special case for 2 since most templates will generate two segments and we'd like it to be efficient
+        val bp = findBreakpoint(segments.head, segments.last, maxWithinReadDistance, maxReadPairInnerDistance)
+        if (bp.isEmpty) NoBreakpoints else bp.toIndexedSeq
```
nh13, line 212: Isn't `.toIndexedSeq` redundant?
```suggestion
        if (bp.isEmpty) NoBreakpoints else bp
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
+      case _     =>
+        val builder = IndexedSeq.newBuilder[BreakpointEvidence]
+        segments.iterator.sliding(2).foreach { case Seq(seg1, seg2) =>
+          findBreakpoint(seg1, seg2, maxWithinReadDistance, maxReadPairInnerDistance).foreach(builder += _)
+        }
+        builder.result()
```
nh13, line 218: Why not?
```suggestion
        segments.iterator.sliding(2).flatMap { case Seq(seg1, seg2) =>
          findBreakpoint(seg1, seg2, maxWithinReadDistance, maxReadPairInnerDistance)
        }.toIndexedSeq
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
+      val ev = if (seg1.origin.isInterRead(seg2.origin)) EvidenceType.ReadPair else EvidenceType.SplitRead
+      Some(BreakpointEvidence(bp, ev))
+    }
+    else {
+      None
     }
   }
 
-  /** Determines if two segments are evidence of an SV joining two different contigs.
+  /** Determines if two segments are evidence of breakpoint joining two different contigs.
```
nh13, line 239:
```suggestion
  /** Determines if two segments are evidence of a breakpoint joining two different contigs.
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
-      Some(BreakpointEvidence(from=seg1, into=seg2, evidence=evidence))
-    }
+    r1.refIndex != r2.refIndex
   }
 
-  /** Determines if the primary mappings are from a read pair are evidence of an SV joining two different regions from
-   * the same contig.
+  /** Determines if the two segments are provide evidence of a breakpoint joining two different regions from
+   * the same contig. Returns true if:
+   *   - the two segments overlap (implying some kind of duplication)
```
nh13, line 252: I am not sure I agree with this by default.  What about short inserts?  How do we control this?

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -121,157 +120,166 @@ class SvPileup
...
+      val innerDistance = {
+        if (seg1.range.start <= seg2.range.start) seg2.range.start - seg1.range.end
+        else                                      seg1.range.start - seg2.range.end
       }
-    }
-    else {
-      if (seg1.positiveStrand == seg2.positiveStrand) None
-      else Some(BreakpointEvidence(from=seg1, into=seg2, evidence=SplitReadOppositeStrand))
+
+      innerDistance > maxDistance
```
nh13, line 282: I don't find the above easier to read, due to mixing of the nested conditionals all being part of the return expression.  Why not the following?  It makes commenting easier

```scala
if (seg1.positiveStrand != seg2.positiveStrand) true // opposite strands
else if (seg1.positiveStrand && seg2.range.start < seg1.range.end) true // one segment before the other
else if (!seg1.positiveStrand && seg1.range.start < seg2.range.end) true // one segment before the other
else {  // check inner distance
  val innerDistance = {
    if (seg1.range.start <= seg2.range.start) seg2.range.start - seg1.range.end
    else                                      seg1.range.start - seg2.range.end
  }
  innerDistance > maxDistance
}
```

nh13 (approved):
Minor comments on syntax, but agree with all the functional changes you've done.

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -147,43 +146,44 @@ object AlignedSegment extends LazyLogging {
...
+    val r1Segments = (r1Primary, template.r1Supplementals) match {
+      case (Some(pri), supps) if supps.isEmpty => IndexedSeq(AlignedSegment(pri))
+      case (Some(pri), supps)                  => segmentsFrom(pri, supps, minUniqueBasesToAdd)
+      case (None     , _    )                  => NoSegments
+    }
+
+    val r2Segments = (r2Primary, template.r2Supplementals) match {
+      case (Some(pri), supps) if supps.isEmpty => IndexedSeq(AlignedSegment(pri))
+      case (Some(pri), supps)                  => segmentsFrom(pri, supps, minUniqueBasesToAdd)
+      case (None     , _    )                  => NoSegments
```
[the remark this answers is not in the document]
tfenne: I think when I first refactored to these matches `supps` was still an iterator.

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -147,43 +146,44 @@ object AlignedSegment extends LazyLogging {
...
     }
+
+    if (r1Segments.isEmpty)
+      r2Segments
+    else if (r2Segments.isEmpty)
+      r1Segments
+    else
+      // Assume the library should be FR, reverse the R2 segments and flip their strand so that they're
+      // in the same order and orientation as R1 segments
+      mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments.reverse.map(b => b.copy(positiveStrand = !b.positiveStrand)))
```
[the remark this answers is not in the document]
tfenne: Converted it to a match instead.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
tfenne: `bp` is an Option which does not implement `IndexedSeq`.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
tfenne: I'll add a test, but in that case the two reads will generate two segments, which overlap and will get merged into a single large segment, so we won't see two overlapping segments.

tfenne: Added test, works as expected, added a comment here too.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
tfenne: Fine 😝

PR #6, opened 2022-02-16
Add support for auto-generating docs

nh13:


nh13:
Commit [d3fdeca](https://github.com/fulcrumgenomics/fgsv/pull/6/commits/d3fdeca01b57386f7c85057cfcf31a3546476018) shows the auto-update!

tfenne (approved):
Though is there any way to not copy/paste the doclet around?

nh13:
@tfenne we could refactor the metrics doclet in fgbio significantly.  It may be the case that some packages like fgsv would need to extend it though.  I am not going to do that in this PR.

PR #7, opened 2022-02-18
More fixups of SvPileup; transition to using Metrics class; cleanup o…

tfenne:
…f FilterAndMerge

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/sv/tools/FilterAndMerge.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/tools/FilterAndMerge.scala
@@ -1,17 +1,15 @@
 package com.fulcrumgenomics.sv.tools
 
-import com.fulcrumgenomics.FgBioDef.{BetterBufferedIteratorScalaWrapper, FilePath, PathToSequenceDictionary}
+import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.commons.collection.BetterBufferedIterator
-import com.fulcrumgenomics.commons.util.{DelimitedDataParser, NumericCounter, Row, SimpleCounter}
+import com.fulcrumgenomics.commons.util.{NumericCounter, SimpleCounter}
 import com.fulcrumgenomics.fasta.SequenceDictionary
-import com.fulcrumgenomics.sv.cmdline.{ClpGroups, SvTool}
 import com.fulcrumgenomics.sopt.{arg, clp}
-import com.fulcrumgenomics.sv.EvidenceType
+import com.fulcrumgenomics.sv.cmdline.{ClpGroups, SvTool}
+import com.fulcrumgenomics.sv.{BreakpointPileup, EvidenceType}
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/sv/tools/FilterAndMerge.scala
```diff
@@ -33,166 +31,106 @@ class FilterAndMerge
...
-      .sortBy { p =>
-        (dict(p.left_contig).index, dict(p.right_contig).index, p.left_start, p.left_end, p.right_start, p.right_end)
-      }
-      .foreach { merged =>
-        val values = MergedPileup.values(merged)
-        writer.write(values.mkString("\t"))
-        writer.write('\n')
-      }
-    writer.close()
+    Metric.write(output, results.result().sorted)
```
nh13, line 52: changes the sort :/

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -91,35 +99,33 @@ class SvPileup
...
-      .sortWith(Breakpoint.PairedOrdering.lt)
+  /** Write the breakpoints to a file. */
+  private def writeBreakpoints(path: FilePath,
+                               tracker: BreakpointTracker,
+                               dict: SequenceDictionary): Unit = {
+    val writer      = Metric.writer[BreakpointPileup](path)
+    val breakpoints = tracker.breakpoints.toIndexedSeq.sortWith(Breakpoint.PairedOrdering.lt)
 
     breakpoints.foreach { bp =>
       val info         = tracker(bp)
```
nh13, line 110:
```suggestion
      val info       = tracker(bp)
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -91,35 +99,33 @@ class SvPileup
...
-      val id           = info.id
-      val leftRefName  = dict(bp.leftRefIndex).name
-      val rightRefName = dict(bp.rightRefIndex).name
-
-      val values = Iterator(
-        id, leftRefName, bp.leftPos, toStrand(bp.leftPositive), rightRefName, bp.rightPos, toStrand(bp.rightPositive),
-        info.splitRead, info.readPair, info.total
+      val metric       = BreakpointPileup(
```
nh13, line 111:
```suggestion
      val metric = BreakpointPileup(
```

nh13 (approved):

src/main/scala/com/fulcrumgenomics/sv/tools/FilterAndMerge.scala
tfenne: Well the old sort involved the `sameStrand` which I don't think makes too much sense any more.  If you have a different preference for how to order fields I'm game - I thought I did this in a way that would give you the closest to what you had before (it's using the ordering on the new metric class, not Breakpoint).

PR #8, opened 2022-03-23
SvPileup better identifies reads with SV evidence

nh13:
Previously, all reads in a template would be annotated with the SV
evidence if any of the reads in the template exhibited SV evidence.
Now, only those reads that show evidence of a specific breakpoint
are annoated.

### DIFF
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -19,7 +19,7 @@ Two output files will be created:
 2. `<output-prefix>.bam`: a SAM/BAM file containing reads that contain SV breakpoint evidence annotated with SAM
   tag.
 
-The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
+The `be` SAM tag contains a comma-delimited list of breakends to which a given read belongs.  Each element is
 a semi-colon delimited, with four fields:
 
 1. The unique breakpoint identifier (same identifier found in the tab-delimited output).
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -19,7 +19,7 @@ Two output files will be created:
 2. `<output-prefix>.bam`: a SAM/BAM file containing reads that contain SV breakpoint evidence annotated with SAM
   tag.
 
-The `be` SAM tag contains a comma-delimited list of breakends to which a given read belongs.  Each element is
+The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
 a semi-colon delimited, with four fields:
 
 1. The unique breakpoint identifier (same identifier found in the tab-delimited output).
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -20,14 +20,16 @@ Two output files will be created:
   tag.
 
 The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
-a semi-colon delimited, with three fields:
+a semi-colon delimited, with four fields:
 
 1. The unique breakpoint identifier (same identifier found in the tab-delimited output).
-2. Either "from" or "into", such that when traversing the breakpoint would read through "from" and then into
+2. Either "left" or "right, corresponding to if the read shows evidence of the genomic left or right side of the
+   breakpoint as found in the breakpoint file (i.e. `left_pos` or `right_pos`).
+3. Either "from" or "into", such that when traversing the breakpoint would read through "from" and then into
    "into" in the sequencing order of the read pair.  For a split-read alignment, the "from" contains the aligned
    portion of the read that comes from earlier in the read in sequencing order.  For an alignment of a read-pair
    spanning the breakpoint, then "from" should be read-one of the pair and "into" should be read-two of the pair.
-3. The type of breakpoint evidence: either "split_read" for observations of an aligned segment of a single read
+4. The type of breakpoint evidence: either "split_read" for observations of an aligned segment of a single read
    with split alignments, or "read_pair" for observations _between_ reads in a read pair.
 
 ## Algorithm Overview
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -9,12 +9,45 @@ title: SvPileup
 
 Collates a pileup of putative structural variant supporting reads.
 
+## Outputs
+
 Two output files will be created:
 
-1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpiont event.
+1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpoint event.  The returned
+   breakpoint will be canonicalized such that the "left" side of the breakpoint will have the lower (or equal to)
+   position on the genome vs. the "right"s side.
 2. `<output-prefix>.bam`: a SAM/BAM file containing reads that contain SV breakpoint evidence annotated with SAM
-  tags.  The `ev` SAM tag lists the type of evidence found, while the `be` list the unique breakpoint identifier
-  output in (1) above.
+  tag.
+
+The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
+a semi-colon delimited, with three fields:
+
+1. The unique breakpoint identifier (same identifier found in the tab-delimited output).
+2. Either "from" or "into", such that when traversing the breakpoint would read through "from" and then into
+   "into" in the sequencing order of the read pair.  For a split-read alignment, the "from" contains the aligned
+   portion of the read that comes from earlier in the read in sequencing order.  For an alignment of a read-pair
+   spanning the breakpoint, then "from" should be read-one of the pair and "into" should be read-two of the pair.
+3. The type of breakpoint evidence: either "split_read" for observations of an aligned segment of a single read
+   with split alignments, or "read_pair" for observations _between_ reads in a read pair.
+
+## Algorithm Overview
+
+Putative breakpoints are identified by examining the alignments for each template. The alignments are transformed
+into aligned segments in the order they were sequenced.  Each aligned segment represents the full genomic span of
+the mapped bases.  This is performed first for the primary alignments.  Next, supplementary alignments are added
+only if they map read bases that have not been previously covered by other alignments (see
+`--min-unique-bases-to-add`).  This is iteratively performed until supplementary alignments have been exhausted.
+
+Next, aligned segments that have overlapping genomic mapped bases are merged into a single aligned
+segment.  In this case, the two or more read mappings merged are associated with either the left side or right
+side of that aligned segment, controlled by examining how close to the end of the new aligned segment the given
+read mapping occurs (see `--slop` option).  This used to identify which reads traverse "from" and "into" the
+breakpoint as described above.
+
+Finally, pairs of adjacent aligned segments are examined for evidence of a breakpoint, such that genomic distance
+between them beyond either `--max-read-pair-inner-distance` for aligned segments from different read pairs, or
+`--max-aligned-segment-inner-distance` for aligned segments from the same read in a pair (i.e. split-read mapping).
+Split read evidence will be returned in favor of across-read-pair evidence when both are present.
 
 ## Arguments
 
@@ -27,4 +60,5 @@ Two output files will be created:
 |min-primary-mapping-quality|q|Int|The minimum mapping quality for primary alignments|Optional|1|30|
 |min-supplementary-mapping-quality|Q|Int|The minimum mapping quality for supplementary alignments|Optional|1|18|
 |min-unique-bases-to-add|b|Int|The minimum # of uncovered query bases needed to add a supplemental alignment|Optional|1|20|
+|slop|s|Int|The number of bases of slop to allow when determining which records to track for the left or right side of an aligned segment when merging segments.|Optional|1|5|
 
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -32,6 +32,27 @@ a semi-colon delimited, with four fields:
 4. The type of breakpoint evidence: either "split_read" for observations of an aligned segment of a single read
    with split alignments, or "read_pair" for observations _between_ reads in a read pair.
 
+## Example output
+
+The following shows two breakpoints:
+
+```
+id left_contig left_pos left_strand right_contig right_pos right_strand split_reads read_pairs total
+ 1        chr1      100           +         chr2       200            -           1          0     1
+ 2        chr2      150           -         chr3       500            +           1          0     1
+```
+
+Consider a single fragment read that maps across both the above two breakpoints, so has three split-read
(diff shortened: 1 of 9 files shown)


src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -89,9 +89,10 @@ class SvPileup
                                  tracker: BreakpointTracker,
                                  writer: SamWriter): Unit = {
     if (evidences.nonEmpty) {
-      val bps = evidences.map(e => tracker(e.breakpoint).id).toSet.toSeq.sorted.mkString(",")
-      val evs = evidences.map(_.evidence.snakeName).mkString(",")
       template.allReads.foreach { rec =>
+        val readEvs = evidences.filter(_.recs.contains(rec))
+        val bps     = readEvs.map(e => tracker(e.breakpoint).id).toSet.toSeq.sorted.mkString(",")
+        val evs     = readEvs.map(_.evidence.snakeName).mkString(",")
```
nh13, line 95: rather than adding the tags to all reads, now only the reads that observed a given break point

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -231,9 +232,8 @@ object SvPileup extends LazyLogging {
     if (isInterContigBreakpoint(seg1, seg2) ||
         isIntraContigBreakpoint(seg1, seg2, maxWithinReadDistance, maxReadPairInnerDistance)
     ) {
-      val bp = Breakpoint(seg1, seg2)
       val ev = if (seg1.origin.isInterRead(seg2.origin)) EvidenceType.ReadPair else EvidenceType.SplitRead
-      Some(BreakpointEvidence(bp, ev))
+      Some(BreakpointEvidence(from=seg1, into=seg2, evidence=ev))
```
nh13, line 236: simplify by just using the already present apply method for BreakpointEvidence

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -61,10 +61,10 @@ class SvPileup
...
         // Find the breakpoints
         val evidences = findBreakpoints(
-          template                       = template,
-          maxWithinReadDistance          = maxAlignedSegmentInnerDistance,
-          maxReadPairInnerDistance       = maxReadPairInnerDistance,
-          minUniqueBasesToAdd            = minUniqueBasesToAdd,
+          template                 = template,
+          maxWithinReadDistance    = maxAlignedSegmentInnerDistance,
+          maxReadPairInnerDistance = maxReadPairInnerDistance,
+          minUniqueBasesToAdd      = minUniqueBasesToAdd,
```
nh13, line 67: just alignment

src/test/scala/com/fulcrumgenomics/sv/AlignedSegmentTest.scala
```diff
@@ -18,58 +18,63 @@ class AlignedSegmentTest extends UnitSpec {
   "AlignedSegment" should "be built from a SamRecord" in {
     implicit val builder: SamBuilder = new SamBuilder()
 
+    def test(rec: SamRecord, expected: AlignedSegment): Unit = {
+      val actual = AlignedSegment(rec)
+      actual shouldBe expected.copy(recs=IndexedSeq(rec))
+    }
+
```
nh13, line 25: Lots of changes here, but all are just to accommodate having the `SamRecord` tracked in an `AlignedSegment`.   No new tests.

src/test/scala/com/fulcrumgenomics/sv/UnitSpec.scala
```diff
@@ -1,6 +1,24 @@
...
+import com.fulcrumgenomics.FgBioDef.PathToBam
+import com.fulcrumgenomics.bam.api.{SamRecord, SamSource}
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.{Matchers, OptionValues}
+
+import java.nio.file.{Files, Path}
 
 /** Base class for unit tests. */
-class UnitSpec extends FlatSpec with Matchers with OptionValues {}
+trait UnitSpec extends AnyFlatSpec with Matchers with OptionValues {
```
nh13, line 11: we should really make `fgbio`'s `UnitSpec` public instead of copying everywhere

src/test/scala/com/fulcrumgenomics/sv/tools/SvPileupTest.scala
```diff
@@ -1,12 +1,15 @@
 package com.fulcrumgenomics.sv.tools
 
+import com.fulcrumgenomics.FgBioDef.{FilePath, PathPrefix, PathToBam}
```
nh13, line 3: 1. Updated tests to test that we tracked the `SamRecord`(s) correctly
2. Added an end-to-end test to check the SAM tags are added to the correct records (not all within a template)

docs/tools/index.md
```diff
@@ -4,7 +4,7 @@ title: fgsv tools
 
 # fgsv tools
 
-The following tools are available in fgsv version 20220218-d0eb31d.
+The following tools are available in fgsv version 20220323-9e90d4c.
```
tfenne, line 7: Should we just say "the current version of fgsv" since the commit hash will change on squash?

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -19,13 +19,16 @@ import scala.collection.{BitSet, mutable}
...
+ *                       aligned segments were merged, and be empty if not tracking.
  */
 case class AlignedSegment(origin: SegmentOrigin,
                           private val readStart: Int,
                           private val readEnd: Int,
                           positiveStrand: Boolean,
                           private val cigar: Cigar,
-                          range: GenomicRange) {
+                          range: GenomicRange,
+                          recs: Seq[SamRecord] = IndexedSeq.empty) {
```
tfenne, line 31: Could make this `Iterable[SamRecord]` and then `Some(rec)` and `None` are valid values.

src/main/scala/com/fulcrumgenomics/sv/BreakpointEvidence.scala
```diff
@@ -26,13 +28,8 @@ object BreakpointEvidence {
...
-case class BreakpointEvidence(breakpoint: Breakpoint, evidence: EvidenceType) extends Ordered[BreakpointEvidence] {
-
-  /** Defines an ordering over breakpoints, first by genomic coordinates of the left then right end, then evidence type. */
-  def compare(that: BreakpointEvidence): Int = {
-    var result              = this.breakpoint.compare(that.breakpoint)
-    if (result == 0) result = this.evidence.compare(that.evidence)
-    result
-  }
-}
+case class BreakpointEvidence(breakpoint: Breakpoint,
```
tfenne, line 33: Why remove the `Ordered[]`?

src/test/scala/com/fulcrumgenomics/sv/UnitSpec.scala
tfenne: If you can figure out a way to do that without having to add all the testing libraries to the non-test dependencies n each project I'm all for it.  Likely we just need to make an `fgscalatest` project and put stuff there so we can depened on it only in test configurations.

tfenne (commented):
I think we need to convert to tracking `leftRecs` and `rightRecs` on the AlignedSegment as we discussed in slack, and then track those through when merging.

docs/tools/index.md
nh13: This is all done automatically by `BuildToolDocs`, annoying but happy for someone to make it better.

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
nh13: leaving it as is as it breaks tests even though we have the same contents in the iterable and I don't want to spend the time figuring it out.

src/main/scala/com/fulcrumgenomics/sv/BreakpointEvidence.scala
nh13: It was never being used

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -21,12 +21,45 @@ import scala.collection.mutable
   """
     |Collates a pileup of putative structural variant supporting reads.
     |
+    |## Outputs
```
nh13, line 24: this whole usage was really tough to write, it could use a thorough "not me" read through

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -21,12 +21,45 @@ import scala.collection.mutable
...
+    |
+    |Next, aligned segments that have overlapping genomic mapped bases are merged into a single aligned
+    |segment.  In this case, the two or more read mappings merged are associated with either the left side or right
+    |side of that aligned segment, controlled by examining how close to the end of the new aligned segment the given
+    |read mapping occurs (see `--slop` option).  This used to identify which reads traverse "from" and "into" the
+    |breakpoint as described above.
+    |
+    |Finally, pairs of adjacent aligned segments are examined for evidence of a breakpoint, such that genomic distance
+    |between them beyond either `--max-read-pair-inner-distance` for aligned segments from different read pairs, or
+    |--max-aligned-segment-inner-distance` for aligned segments from the same read in a pair (i.e. split-read mapping).
```
nh13, line 61:
```suggestion
    |`--max-aligned-segment-inner-distance` for aligned segments from the same read in a pair (i.e. split-read mapping).
```

src/test/scala/com/fulcrumgenomics/sv/tools/SvPileupTest.scala
```diff
@@ -361,4 +393,76 @@ class SvPileupTest extends UnitSpec {
...
+      }
+    }
+
+    val fromTag = s"0;from;${SplitRead.snakeName}"
+    val intoTag = f"0;into;${SplitRead.snakeName}"
+
+//    // r1Half2 is not annotated, since it is superseded by r1Half1 -> fullR2
+//    SamPairUtil.setMateInfo(r1Half1.asSam, fullR2.asSam, true)
+//    SamPairUtil.setMateInformationOnSupplementalAlignment(r1Half2.asSam, fullR2.asSam, true)
+//    test(Seq(r1Half2, r1Half1, fullR2), Seq(fromTag, "", intoTag))
```
nh13, line 455:
```suggestion
    // r1Half2 is not annotated, since it is superseded by r1Half1 -> fullR2
    SamPairUtil.setMateInfo(r1Half1.asSam, fullR2.asSam, true)
    SamPairUtil.setMateInformationOnSupplementalAlignment(r1Half2.asSam, fullR2.asSam, true)
    test(Seq(r1Half2, r1Half1, fullR2), Seq(fromTag, "", intoTag))
```

docs/tools/SvPileup.md
```diff
@@ -9,12 +9,47 @@ title: SvPileup
...
-1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpiont event.
+1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpoint event.  The returned
+   breakpoint will be canonicalized such that the "left" side of the breakpoint will have the lower (or equal to)
+   position on the genome vs. the "right"s side.
 2. `<output-prefix>.bam`: a SAM/BAM file containing reads that contain SV breakpoint evidence annotated with SAM
-  tags.  The `ev` SAM tag lists the type of evidence found, while the `be` list the unique breakpoint identifier
-  output in (1) above.
+  tag.
+
+The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
```
tfenne, line 22:
```suggestion
The `be` SAM tag contains a comma-delimited list of breakends to which a given read belongs.  Each element is
```

docs/tools/SvPileup.md
```diff
@@ -9,12 +9,47 @@ title: SvPileup
...
+a semi-colon delimited, with four fields:
```
tfenne, line 23: Maybe give an example of a read that supports two breakends (is that really a thing?) just to make it clear how `,` and `;` are used.

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -19,13 +19,21 @@ import scala.collection.{BitSet, mutable}
...
  */
 case class AlignedSegment(origin: SegmentOrigin,
                           private val readStart: Int,
                           private val readEnd: Int,
                           positiveStrand: Boolean,
                           private val cigar: Cigar,
-                          range: GenomicRange) {
+                          range: GenomicRange,
+                          left: Seq[SamRecord] = IndexedSeq.empty,
+                          right: Seq[SamRecord] = IndexedSeq.empty) {
```
tfenne, line 36: `IndexedSeq.empty` is actually a function.  Would be best to just store an empty `IndexedSeq` in the companion object, e.g. `val NoReads = IndexedSeq.empty[SamRecord]` and then refer to it here.

src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
```diff
@@ -76,7 +103,16 @@ object AlignedSegment extends LazyLogging {
...
-    AlignedSegment(origin = SegmentOrigin(rec), readStart = start, readEnd = end, positiveStrand = rec.positiveStrand, cigar = rec.cigar, range = range)
+    AlignedSegment(
+      origin         = SegmentOrigin(rec),
+      readStart      = start,
+      readEnd        = end,
+      positiveStrand = rec.positiveStrand,
+      cigar          = rec.cigar,
+      range          = range,
+      left           = IndexedSeq(rec),
+      right          = IndexedSeq(rec)
```
tfenne, line 114: Make the IndexedSeq once outside the apply method and use it twice?

src/main/scala/com/fulcrumgenomics/sv/BreakpointEvidence.scala
```diff
@@ -26,13 +36,12 @@ object BreakpointEvidence {
...
+                              evidence: EvidenceType,
+                              from: Set[SamRecord] = Set.empty,
+                              into: Set[SamRecord] = Set.empty,
+                              fromIsLeft: Boolean = true)
```
tfenne, line 47: I want a better name for this, but have no suggestion for a better name :/

tfenne (approved):
A few minor comments, merge when ready.

src/main/scala/com/fulcrumgenomics/sv/BreakpointEvidence.scala
nh13: `lateral`?  To mean that left and from are on the same side.  Leaving for now since I don't either.

PR #9, opened 2022-03-24
Progress logger can count templates with no primary alignment

pamelarussell:


src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -56,7 +56,7 @@ class SvPileup
     val tracker   = new BreakpointTracker()
 
     Bams.templateIterator(source)
-      .tapEach(t => progress.record(t.r1.getOrElse(t.r2.get)))
+      .tapEach(_ => progress.record())
```
nh13, line 59:
```suggestion
      .tapEach(t => progress.record(t.allReads.head))
```

nh13 (changes requested):
I'd question why we have a template with primary alignment, that seems like an issue with the input data.

pamelarussell:
> I'd question why we have a template with primary alignment, that seems like an issue with the input data.

This can happen if you filter a bam file on some criteria and lose the primary alignment of a template.

nh13 (approved):

PR #10, opened 2022-03-31
AggregateSvPileup initial implementation

pamelarussell:
Tool that merges clusters of related pileups generated by `SvPileup`.

### DIFF
--- a/docs/tools/SvPileup.md
+++ b/docs/tools/SvPileup.md
@@ -9,12 +9,68 @@ title: SvPileup
 
 Collates a pileup of putative structural variant supporting reads.
 
+## Outputs
+
 Two output files will be created:
 
-1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpiont event.
+1. `<output-prefix>.txt`: a tab-delimited file describing SV pileups, one line per breakpoint event.  The returned
+   breakpoint will be canonicalized such that the "left" side of the breakpoint will have the lower (or equal to)
+   position on the genome vs. the "right"s side.
 2. `<output-prefix>.bam`: a SAM/BAM file containing reads that contain SV breakpoint evidence annotated with SAM
-  tags.  The `ev` SAM tag lists the type of evidence found, while the `be` list the unique breakpoint identifier
-  output in (1) above.
+  tag.
+
+The `be` SAM tag contains a comma-delimited list of breakpoints to which a given read belongs.  Each element is
+a semi-colon delimited, with four fields:
+
+1. The unique breakpoint identifier (same identifier found in the tab-delimited output).
+2. Either "left" or "right, corresponding to if the read shows evidence of the genomic left or right side of the
+   breakpoint as found in the breakpoint file (i.e. `left_pos` or `right_pos`).
+3. Either "from" or "into", such that when traversing the breakpoint would read through "from" and then into
+   "into" in the sequencing order of the read pair.  For a split-read alignment, the "from" contains the aligned
+   portion of the read that comes from earlier in the read in sequencing order.  For an alignment of a read-pair
+   spanning the breakpoint, then "from" should be read-one of the pair and "into" should be read-two of the pair.
+4. The type of breakpoint evidence: either "split_read" for observations of an aligned segment of a single read
+   with split alignments, or "read_pair" for observations _between_ reads in a read pair.
+
+## Example output
+
+The following shows two breakpoints:
+
+```
+id left_contig left_pos left_strand right_contig right_pos right_strand split_reads read_pairs total
+ 1        chr1      100           +         chr2       200            -           1          0     1
+ 2        chr2      150           -         chr3       500            +           1          0     1
+```
+
+Consider a single fragment read that maps across both the above two breakpoints, so has three split-read
+alignments.  The first alignment maps on the left side of breakpoint #1, the second alignment maps to both the
+right side of breakpoint #1 and the left-side of breakpoint #2, and the third alignment maps to the right side of
+breakpoint #2. The SAM records would be as follows:
+
+```
+r1    0 chr1  50 60   50M100S ... be:Z:1;left;from;split_read
+r1 2064 chr2 150 60 50S50M50S ... be:Z:1;right;into;split_read,2;left;from;split_read
+r1 2048 chr3 500 60   100S50M ... be:Z:2;right;into;split_read
+```
+
+## Algorithm Overview
+
+Putative breakpoints are identified by examining the alignments for each template. The alignments are transformed
+into aligned segments in the order they were sequenced.  Each aligned segment represents the full genomic span of
+the mapped bases.  This is performed first for the primary alignments.  Next, supplementary alignments are added
+only if they map read bases that have not been previously covered by other alignments (see
+`--min-unique-bases-to-add`).  This is iteratively performed until supplementary alignments have been exhausted.
+
+Next, aligned segments that have overlapping genomic mapped bases are merged into a single aligned
+segment.  In this case, the two or more read mappings merged are associated with either the left side or right
+side of that aligned segment, controlled by examining how close to the end of the new aligned segment the given
+read mapping occurs (see `--slop` option).  This used to identify which reads traverse "from" and "into" the
+breakpoint as described above.
+
+Finally, pairs of adjacent aligned segments are examined for evidence of a breakpoint, such that genomic distance
+between them beyond either `--max-read-pair-inner-distance` for aligned segments from different read pairs, or
+`--max-aligned-segment-inner-distance` for aligned segments from the same read in a pair (i.e. split-read mapping).
+Split read evidence will be returned in favor of across-read-pair evidence when both are present.
 
 ## Arguments
 
@@ -27,4 +83,5 @@ Two output files will be created:
 |min-primary-mapping-quality|q|Int|The minimum mapping quality for primary alignments|Optional|1|30|
 |min-supplementary-mapping-quality|Q|Int|The minimum mapping quality for supplementary alignments|Optional|1|18|
 |min-unique-bases-to-add|b|Int|The minimum # of uncovered query bases needed to add a supplemental alignment|Optional|1|20|
+|slop|s|Int|The number of bases of slop to allow when determining which records to track for the left or right side of an aligned segment when merging segments.|Optional|1|5|
 
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -5,6 +5,7 @@ title: fgsv tools
 # fgsv tools
 
 The following tools are available in fgsv version 20220331-f977739.
+
 ## All tools
 
 All tools.
--- a/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/AlignedSegment.scala
@@ -19,13 +19,21 @@ import scala.collection.{BitSet, mutable}
  * @param positiveStrand true if the read was mapped to the positive strand, false otherwise
  * @param cigar          the cigar **in sequencing order**
  * @param range          the genomic range to which this alignment maps
+ * @param left           the records from which this aligned segment originated from and abut the left genomic range of
+ *                       this segment.  May contain multiple if two aligned segments were merged, and be empty if not
+ *                       tracking.
+ * @param right          the records from which this aligned segment originated from and abut the left genomic range of
+ *                       this segment.  May contain multiple if two aligned segments were merged, and be empty if not
+ *                       tracking.
  */
 case class AlignedSegment(origin: SegmentOrigin,
                           private val readStart: Int,
                           private val readEnd: Int,
                           positiveStrand: Boolean,
                           private val cigar: Cigar,
-                          range: GenomicRange) {
+                          range: GenomicRange,
+                          left: Seq[SamRecord] = AlignedSegment.NoReads,
+                          right: Seq[SamRecord] = AlignedSegment.NoReads) {
   require(0 < readStart)
   require(readStart <= readEnd)
 
@@ -42,17 +50,38 @@ case class AlignedSegment(origin: SegmentOrigin,
   /** Merges the other segment with itself, returning a new alignment segment.  The two segments must overlap on the genome
    * and map to teh same strand.  If merging segments from different reads (i.e. R1 and R2), the segment origin will be set
    * to both.  The resulting cigar will be empty, and the start and end offset in the read will both be 1.
-   * */
-  def merge(other: AlignedSegment): AlignedSegment = {
+   *
+   * @param other the other segment to merge
+   * @param slop the existing tracked records in this or other will be tracked in the new alignment based on being within
+   *             the given slop of the new segments' left-most or right-most position respectively.
+   * @return
+   */
+  def merge(other: AlignedSegment, slop: Int = 0): AlignedSegment = {
     require(this.overlapsWithStrand(other))
-    val range = this.range.union(other.range)
+    val range  = this.range.union(other.range)
     val origin = if (this.origin == other.origin) this.origin else SegmentOrigin.Both
-    AlignedSegment(origin=origin, readStart=1, readEnd=1, positiveStrand=this.positiveStrand, cigar=Cigar.empty, range=range)
+    val left   = (math.abs(this.range.start - range.start) <= slop, math.abs(other.range.start - range.start) <= slop) match {
+      case (true, true)  => this.left ++ other.left
+      case (true, false) => this.left
+      case (false, true) => other.left
+      case _             => throw new IllegalStateException("Bug: no left end tracked alignments")
+    }
+    val right = (math.abs(this.range.end - range.end) <= slop, math.abs(other.range.end - range.end) <= slop) match {
+      case (true, true)  => this.right ++ other.right
+      case (true, false) => this.right
+      case (false, true) => other.right
+      case _             => throw new IllegalStateException("Bug: no right end tracked alignments")
+    }
+    AlignedSegment(
+      origin=origin, readStart=1, readEnd=1, positiveStrand=this.positiveStrand, cigar=Cigar.empty, range=range, left=left, right=right
+    )
   }
 }
 
 object AlignedSegment extends LazyLogging {
   private val NoSegments: IndexedSeq[AlignedSegment] = IndexedSeq.empty
+
+  private val NoReads: IndexedSeq[SamRecord] = IndexedSeq.empty[SamRecord]
 
   /** Builds an alignment segment from a [[SamRecord]]
    *
@@ -76,7 +105,18 @@ object AlignedSegment extends LazyLogging {
       else                    (trailingClipping + 1, trailingClipping + middle)
     }
 
-    AlignedSegment(origin = SegmentOrigin(rec), readStart = start, readEnd = end, positiveStrand = rec.positiveStrand, cigar = rec.cigar, range = range)
+    val recs = IndexedSeq(rec)
+
+    AlignedSegment(
+      origin         = SegmentOrigin(rec),
+      readStart      = start,
+      readEnd        = end,
+      positiveStrand = rec.positiveStrand,
+      cigar          = rec.cigar,
+      range          = range,
+      left           = recs,
+      right          = recs
+    )
   }
 
   /** Builds [[AlignedSegment]]s for the given alignments for a given read, one per record. The segments returned
@@ -155,11 +195,13 @@ object AlignedSegment extends LazyLogging {
    * it is added to the set.  This is performed until no more supplementary alignments remain.
    * 3. The segments for R1 and R2 produced in step (2) are merged with the procedure described in [[mergeAlignedSegments()]].
    *
-   * @param template                       the template from which alignment segments are to be produced
-   * @param minUniqueBasesToAdd            the minimum # of new bases that a supplementary alignment must map to keep it in the
-   *                                       iterative procedure described above.
+   * @param template            the template from which alignment segments are to be produced
+   * @param minUniqueBasesToAdd the minimum # of new bases that a supplementary alignment must map to keep it in the
+   *                            iterative procedure described above.
+   * @param slop                the number of bases of slop to allow when determining which records to track for the
+   *                            left or right side of an aligned segment when merging segments
    */
-  def segmentsFrom(template: Template, minUniqueBasesToAdd: Int): IndexedSeq[AlignedSegment] = {
+  def segmentsFrom(template: Template, minUniqueBasesToAdd: Int, slop: Int = 0): IndexedSeq[AlignedSegment] = {
     val r1Primary = template.r1.filter(_.mapped)
     val r2Primary = template.r2.filter(_.mapped)
     require(r1Primary.nonEmpty || r2Primary.nonEmpty, s"${template.name} did not have a primary R1 or R2 mapped.")
@@ -176,13 +218,17 @@ object AlignedSegment extends LazyLogging {
       case (None     , _    ) => NoSegments
     }
 
+    // Merge segments if necessary
     (r1Segments, r2Segments) match {
       case (segs, IndexedSeq()) => segs
       case (IndexedSeq(), segs) => segs
-      case (s1, s2) =>
+      case _                    =>
         // Assume the library should be FR, reverse the R2 segments and flip their strand so that they're
         // in the same order and orientation as R1 segments
-        mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments.reverse.map(b => b.copy(positiveStrand = !b.positiveStrand)))
+        mergeAlignedSegments(
+          r1Segments = r1Segments,
+          r2Segments = r2Segments.reverse.map(b => b.copy(positiveStrand = !b.positiveStrand)),
+          slop       = slop)
     }
   }
 
@@ -201,12 +247,15 @@ object AlignedSegment extends LazyLogging {
    * @param r1Segments             the alignment segments from read one sorted ascending the start of the read in sequencing order.
    * @param r2Segments             the alignment segments from read two sorted ascending the start of the read in sequencing order.
    * @param numOverlappingSegments the # of overlapping segments to examine
+   * @param slop                   the number of bases of slop to allow when determining which records to track for the
+   *                               left or right side of an aligned segment when merging segments
    * @return
    */
   @tailrec
   def mergeAlignedSegments(r1Segments: IndexedSeq[AlignedSegment],
                            r2Segments: IndexedSeq[AlignedSegment],
-                           numOverlappingSegments: Int = 1): IndexedSeq[AlignedSegment] = {
+                           numOverlappingSegments: Int = 1,
+                           slop: Int = 0): IndexedSeq[AlignedSegment] = {
     if (numOverlappingSegments > r1Segments.length || numOverlappingSegments > r2Segments.length) {
       r1Segments ++ r2Segments // no overlapping segments
     } else {
@@ -217,11 +266,11 @@ object AlignedSegment extends LazyLogging {
         // merge and return
         val leading  = r1Segments.take(r1Segments.length - numOverlappingSegments)
         val trailing = r2Segments.takeRight(r2Segments.length - numOverlappingSegments)
-        val middle   = r1SubSegments.zip(r2SubSegments).map { case (seg1, seg2) => seg1.merge(seg2) }
+        val middle   = r1SubSegments.zip(r2SubSegments).map { case (seg1, seg2) => seg1.merge(seg2, slop=slop) }
         leading ++ middle ++ trailing
       } else {
         // recurse
-        mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments, numOverlappingSegments=numOverlappingSegments + 1)
+        mergeAlignedSegments(r1Segments=r1Segments, r2Segments=r2Segments, numOverlappingSegments=numOverlappingSegments + 1, slop=slop)
       }
     }
   }
--- a/src/main/scala/com/fulcrumgenomics/sv/Breakpoint.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/Breakpoint.scala
@@ -11,10 +11,12 @@ object Breakpoint {
    * segments could be from different reads in a mate pair, this cannot be validated/required()
    * in this method, but violating this assumption will lead to invalid breakpoints.
    *
-   * The returned breakpoint will be canonicalized such that the `left` side of the breakpoint will have the
-   * lower (or equal to) position on the genome vs. the `right` side.
+   * @param into the segment earlier in sequencing order
+   * @param from the segment later in sequencing order
+   * @param canonicalize canonicalize the breakpoint such that the `left` side of the breakpoint will have the lower
+   *                     (or equal to) position on the genome vs. the `right` side.
    */
-  def apply(from: AlignedSegment, into: AlignedSegment): Breakpoint = {
+  def apply(from: AlignedSegment, into: AlignedSegment, canonicalize: Boolean = true): Breakpoint = {
     val bp = Breakpoint(
       leftRefIndex  = from.range.refIndex,
       leftPos       = if (from.positiveStrand) from.range.end else from.range.start,
@@ -24,7 +26,7 @@ object Breakpoint {
       rightPositive = into.positiveStrand
     )
 
-    if (bp.isCanonical) bp else bp.reversed
+    if (!canonicalize || bp.isCanonical) bp else bp.reversed
   }
 
   /**
@@ -90,7 +92,7 @@ case class Breakpoint(leftRefIndex: Int,
     rightPositive = !leftPositive
   )
 
-  /** Returns true if the representation of the breakend is canonical, with the left hand side of the break
+  /** Returns true if the representation of the breakpoint is canonical, with the left hand side of the break
    * earlier on the genome than the right hand side. */
   def isCanonical: Boolean = (leftRefIndex < rightRefIndex) ||
     (leftRefIndex == rightRefIndex && leftPos < rightPos) ||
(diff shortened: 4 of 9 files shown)


src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+    |
+    |The output file is a tab-delimited table with one record per aggregated cluster of pileups. Aggregated
+    |pileups are reported with the minimum and maximum (inclusive) coordinates of all pileups in the cluster, a
+    |possible putative structural variant event type supported by the pileups, and the sum of read support from all
+    |pileups in the cluster.
+    |""")
+class AggregateSvPileup
+(@arg(flag='i', doc="The input text file of pileups generated by SvPileup") input: FilePath,
+ @arg(flag='o', doc="The output file") output: FilePath,
+ @arg(flag='d', doc="Distance threshold below which to cluster breakpoints") maxDist: Int = 100,
```
tfenne, line 42: I wonder if we want a pair of thresholds?  One for merging split-reads, and one for merging paired-end?  I would _think_ it would make sense to have a pretty low threshold (e.g 10-20bp) for differences in split-read positions, but a _much_ larger one (e.g. 250bp) for paired-end/read-pair?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+  /**
+   * Maps each pileup to the set of other pileups within the distance threshold
+   * @param pileups Set of pileups to evaluate for pairwise neighbor relationships
+   * @param maxDist Maximum distance (inclusive) for the left and right coordinates to be considered a neighbor
+   * @return Map of pileup to the set of neighbors
+   */
+  private def pileupToNeighbors(pileups: PileupGroup, maxDist: Int): Map[BreakpointPileup, PileupGroup] = {
+    val map: mutable.Map[BreakpointPileup, PileupGroup] = mutable.Map.from(pileups.map((_, Seq())))
+    for (pileup1 <- pileups; pileup2 <- pileups) {
+      if (pileup1.hashCode() < pileup2.hashCode() && isCluster(pileup1, pileup2, maxDist)) {
```
tfenne, line 87: A couple of thoughts:
1. I think it might be nice to use `pileup1.total >= pilup2.total` here, so that you tend to put things into groups using the _most_ supported breakpoints as the keys
2. Either way, I think you need to use `<=` or >=` here since two pileups could in theory have a hash collision, in which case you wouldn't end up putting then into the map either way
3. To reduce iteration you could a) turn the PileupGroup into an `IndexedSeq` on load/ splitting by chrom/strand, and then to `for (i <- Range(0, pileups.size); j <- Range(i, pileups.size)) {`.  You could also do this using `pileups.tails` but that is marginally more complicated in my view.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+        map.put(pileup1, map(pileup1).appended(pileup2))
+        map.put(pileup2, map(pileup2).appended(pileup1))
```
tfenne, line 89: Might consider putting `Seq.newBuilder` into the maps instead of `Seq` so you can just add records as you go.  Or if you don't care about ordering you could use `Lists` instead (Lists are Seqs) in which case you could:

1. Prepopulate your map with `pileups.map((_, Nil))`
2. Add things here by doing `map.put(pileup1, pileup2 :: map(pileup1))` to avoid all the extra allocation

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+    map.toMap
+  }
+
+  /** Returns true iff the two breakpoints can be clustered, i.e., they have the same left and right chromosomes and
```
tfenne, line 95:
```suggestion
  /** Returns true if the two breakpoints can be clustered, i.e., they have the same left and right chromosomes and
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+
+  /**
+   * Similar to `toClusters()`, but picking up from a partially complete recursive call stack.
+   * @param existingClusters The set of clusters that have already been assembled
+   * @param remainingNeighborGroups The set of remaining neighbor groups, which are all separate from the existing
+   *                                clusters (don't belong with any of the existing clusters)
+   * @return The set of pileup clusters
+   */
+  @tailrec
+  private def finalizeClusters(existingClusters: Seq[PileupGroup],
```
tfenne, line 121: Why not just have these be one method with `existingClusters: Seq[PileupGroup] = Seq()`?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+   * Extracts a complete cluster from the map of pileups to their neighbor sets, and returns the cluster along with the
+   * reduced map of pileup to neighbor set such that no remaining pileups in the map belong in a cluster with the
+   * returned cluster.
+   * @param pileupToNeighbors Map of pileup to its set of neighbors
+   * @return (1) One complete cluster; (2) Remaining map of pileup to neighbor set, all of which are independent of the
+   *         returned cluster.
+   */
+  private def extractOneCluster(pileupToNeighbors: Map[BreakpointPileup, PileupGroup]):
+  (PileupGroup, Map[BreakpointPileup, PileupGroup]) = {
+    extractClusterFor(pileupToNeighbors.keys.head, pileupToNeighbors)
```
tfenne, line 144: Rather than just using `.head` here, how about having it preferentially pick the most supported breakpoint that we have as the next thing to build a cluster around?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+   * @return (1) The complete cluster for the given pileup; (2) Remaining map of pileup to neighbor set, all of which
+   *         are independent of the returned cluster.
+   */
+  def extractClusterFor(pileup: BreakpointPileup, pileupToNeighbors: Map[BreakpointPileup, PileupGroup]):
+  (PileupGroup, Map[BreakpointPileup, PileupGroup]) = {
+    finalizeClusterFor(pileup, Seq(), pileupToNeighbors)
+  }
+
+  /**
+   * Similar to `extractClusterFor()`, but picking up from a partially complete recursive call stack.
```
tfenne, line 162: Again, why not use default argument values to have it be a single function?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+        (neighbors :+ pileup, reducedRemainingGroups))(
+        (currClusterAndRemaining: (PileupGroup, Map[BreakpointPileup, PileupGroup]), nextNeighbor: BreakpointPileup) => {
+          currClusterAndRemaining match {
+            case (currCluster, remaining) =>
+              finalizeClusterFor(nextNeighbor, currCluster, remaining) match {
+                case (expandedCluster, reducedRemaining) => ((currCluster ++ expandedCluster).distinct, reducedRemaining)
+              }
+          }
+        }
+      )
```
tfenne, line 191: I've been staring at this for 5 minutes and am having a really hard time following what it does.  I think the following would help me:

1. More method level doc on exactly what is returned.  I _think_ what is returned is a Map that has any `BreakpointPileup` that is added to the cluster removed as both a key in the returned `Map` and removed from all values in the map?
2. Use of more braces in the code to make it easier to see where blocks start/end
3. More inline comments on exactly what each operation is doing

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+    }
+  }
+
```
tfenne, line 194:
```suggestion
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+/** Contig and strand for the left and right sides of a breakpoint pileup */
+case class BreakpointContigOrientation(left_contig: String, left_strand: Char, right_contig: String, right_strand: Char)
+
+
+object BreakpointContigOrientation {
+
```
tfenne, line 203:
```suggestion
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+  def apply(pileup: BreakpointPileup): BreakpointContigOrientation = {
+    BreakpointContigOrientation(
+      left_contig  = pileup.left_contig,
+      left_strand  = pileup.left_strand,
+      right_contig = pileup.right_contig,
+      right_strand = pileup.right_strand,
+    )
+  }
+
```
tfenne, line 212:
```suggestion
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+/** Type of structural variant supported by a pileup */
+object BreakpointCategory extends Enumeration {
+  type BreakpointCategory = String
+  val POSSIBLE_DELETION = "Possible deletion"
+  val POSSIBLE_INTRA_CONTIG = "Possible intra-contig rearrangement"
+  val POSSIBLE_INTER_CONTIG = "Possible inter-contig rearrangement"
```
tfenne, line 221: No shouting case please!  In scala we use `TitleCase` for constants.

tfenne: I think we can drop the "possible" from the latter two.  I thought we wanted to call the first one possible/likely etc. as we can't be sure it's a deletion and not something else going on, but the latter two are broad enough that they are effectively always correct.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+
+@clp(group=ClpGroups.All, description=
+  """
+    |Merges nearby pileups of reads supporting putative structural variants.
+    |
+    |Takes as input the file of pileups produced by `SvPileup`. That file contains a list of breakpoints, each
+    |consisting of a chromosome, position and strand for each side of the breakpoint, as well as quantified read support
+    |for the breakpoint.
+    |
+    |This tool merges pairs of breakpoints that have their left and right ends on the same chromosomes, the same left
```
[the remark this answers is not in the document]
nh13: Add a "respectively" please, since it sounds like the left chromosome has to be the same as the right chromosome.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+    val pileups: Seq[BreakpointPileup] = Metric.read[BreakpointPileup](input)
+
+    // Group pileups by left and right contig and strand
+    val pileupsByContigStrand: Map[BreakpointContigOrientation, PileupGroup] =
+      pileups.groupMap(BreakpointContigOrientation(_))(identity)
+
+    // Within each contig+strand group, identify all immediate "neighbors" of each pileup (other pileups within the
+    // distance threshold)
+    val pileupToNeighbors: Map[BreakpointContigOrientation, Map[BreakpointPileup, PileupGroup]] =
+      pileupsByContigStrand.view.mapValues(AggregateSvPileup.pileupToNeighbors(_, maxDist)).toMap
```
nh13, line 59: Why not just sort the pileup group by `left_pos` then `right_pos`, then it's easy to linearly iterate to do:

```scala
if (prev.left_pos - cur.left_pos <= maxDist && prev.right_pos - cur.right_pos <= maxDist) {
  // add to current group
} else {
 // finalize previous group, and add the cur to a new one
}
```

what am I missing?

nh13 (changes requested):
@pamelarussell I had an idea for simplifying the tool.  Could you see if you agree or if you can poke a hole in it?  I'll keep reviewing if the latter.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
pamelarussell: @nh13 here's a counterexample:

|id|left_pos|right_pos|
|---|---|---|
|1|1000|5000|
|2|1000|7000|
|3|1001|5000|
|4|1001|7000|

These will sort in the order listed: `1`, `2`, `3`, `4`. You will get no clusters by the method you described. However, there should be two clusters: `1,3` and `2,4`.

nh13: You're right!

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+) extends SvTool with AggregateSvPileupDef {
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    // Read pileups from input file
+    val pileups: Seq[BreakpointPileup] = Metric.read[BreakpointPileup](input)
+
+    // Group pileups by left and right contig and strand
```
nh13, line 52: I am getting a bit confused by having nested maps (i.e. `Map[BreakpointContigOrientation, Map[_, _]]`).  Why not just group by `BreakpointContigOrientation`, then have a loop to do all the work on a `PileupGroup` one by one?

```scala
val writer = Metric.writer[AggregatedBreakpointPileup](output)
pileups.groupMap(BreakpointContigOrientation(_))(identity).foreach { case (orientation, group) =>
  val pileupToNeighbors: MaMap[BreakpointPileup, PileupGroup] = AggregateSvPileup.pileupToNeighbors(group, maxDist)).toMap
  val pileupClusters: Seq[PileupGroup]                        = AggregateSvPileup.toClusters(pileupToNeighbors)
  val aggregatedPileups: Seq[AggregatedBreakpointPileup]      = pileupClusters.map(AggregatedBreakpointPileup.apply)
  aggregatedPileups.foreach(writer.write)
}
writer.close()
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
nh13: I think for (3), the `BreakpointPileup` in `PileupGroup` already have the same `contig` and `strand`, based on how it is being used in the `execute` method (groupby `BreakpointContigOrientation`).  I think this should be stated in the method docs.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+        map.put(pileup2, map(pileup2).appended(pileup1))
+      }
+    }
+    map.toMap
+  }
+
+  /** Returns true iff the two breakpoints can be clustered, i.e., they have the same left and right chromosomes and
+   * strand, and the distance between their left and right coordinates is below the threshold (inclusive) */
+  private def isCluster(bp1: BreakpointPileup, bp2: BreakpointPileup, maxDist: Int): Boolean = {
+    BreakpointContigOrientation(bp1) == BreakpointContigOrientation(bp2) &&
```
nh13, line 98: Since you already grouped by `BreakpointContigOrientation`, shouldn't this always be true?  So can we delete?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,319 @@
...
+    val pileupToNeighbors: Map[BreakpointContigOrientation, Map[BreakpointPileup, PileupGroup]] =
+      pileupsByContigStrand.view.mapValues(AggregateSvPileup.pileupToNeighbors(_, maxDist)).toMap
+
+    // Within each contig+strand group, recursively build (transitive) clusters of neighboring pileups
+    val pileupClusters: Map[BreakpointContigOrientation, Seq[PileupGroup]] =
+      pileupToNeighbors.view.mapValues(AggregateSvPileup.toClusters).toMap
+
+    // Aggregate each cluster
+    val aggregatedPileups: Map[BreakpointContigOrientation, Seq[AggregatedBreakpointPileup]] =
+      pileupClusters.view.mapValues(_.map(AggregatedBreakpointPileup.apply)).toMap
```
nh13, line 67: I am having a really hard time reviewing and following along `toClusters` and friends for what I think should be a really straightforward BFS or DFS search.

If I understand correctly, the algorithm is:

1. group by the same left/right contig+strand
2. for each group, create an adjacency matrix
3. find each connected component in each adjacency matrix
4. for each connected component, create an `AggregatedBreakpointPileup`
5. output

If so, I think the code can be simplified such that after creating the adjacency matrix, we perform a BFS or DFS search within each pileup group

1. Keep a boolean to say if a breakpoint has been "visited", default to false
2. Loop through each breakpoint.  If visited, do nothing.  Else it has not been visited, so add it to the current "cluster", mark as visited, and recurse on all other adjacent breakpoints adding them to the current cluster.
3. Done

nh13 (changes requested):
I'd like to discuss simplifying and making the implementation of `toClusters` and friends a little clearer, for what I think is a BFS/DFS.  See my comments if you agree with the direction I suggested, and we can discuss if you have any questions.

pamelarussell:
Don't re-review yet; I'm considering the DFS suggestion.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
pamelarussell: I'm having trouble designing this algorithm in a way that doesn't require an `extractClusterFor()` method to extract the cluster containing a given representative pileup. Let's discuss live.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+      val pileupToNeighbors: Map[BreakpointPileup, PileupGroup] = AggregateSvPileup.pileupToNeighbors(group, maxDist)
+
+      // Within each contig+strand group, recursively build (transitive) clusters of neighboring pileups
+      val pileupClusters: Seq[PileupGroup] = AggregateSvPileup.toClusters(pileupToNeighbors)
+
+      // Aggregate each cluster
+      val aggregatedPileups: Seq[AggregatedBreakpointPileup] = pileupClusters.map(AggregatedBreakpointPileup.apply)
+
+      // Write to output file
+      aggregatedPileups.foreach(writer.write)
```
nh13, line 64:
```suggestion
      // Aggregate each cluster and write
      pileupClusters.map(AggregatedBreakpointPileup.apply).foreach(writer.write)
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+
+object AggregateSvPileup {
+
+  /** Type alias for a collection of pileups */
+  type PileupGroup = Seq[BreakpointPileup]
+
+  /**
+   * Maps each pileup to the set of other pileups within the distance threshold.
+   *
+   * Note: it is recommended to call this method on a group of pileups that share a common `BreakpointContigOrientation`
```
nh13, line 82: I think this is required, since we're not checking orientation.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+   */
+  @tailrec
+  def toClusters(pileupToNeighbors: Map[BreakpointPileup, PileupGroup],
+                 existingClusters: Seq[PileupGroup] = Seq()): Seq[PileupGroup] = {
+    if (pileupToNeighbors.isEmpty) {
+      existingClusters
+    } else {
+      val (pileupGroup, remaining) = extractClusterFor(pileupToNeighbors.keys.head, pileupToNeighbors)
+      toClusters(remaining, existingClusters :+ pileupGroup)
+    }
```
nh13, line 125: More idiomatic since you want to avoid using `.head` on a list (like avoiding `.get` on an option)
```suggestion
    pileupToNeighbors.keys.headOption match {
      case None         => existingClusters
      case Some(pileup) =>
        val (pileupGroup, remaining) = extractClusterFor(pileup, pileupToNeighbors)
        toClusters(remaining, existingClusters :+ pileupGroup)
    }
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+   * @param existingCluster An existing cluster to add to
+   * @return (1) The complete cluster for the given pileup; (2) Reduced version of `pileupToNeighbors` with all pileups
+   *         in the returned cluster having been removed from both keys and values.
+   */
+  def extractClusterFor(pileup: BreakpointPileup, pileupToNeighbors: Map[BreakpointPileup, PileupGroup],
+                        existingCluster: PileupGroup = Seq()): (PileupGroup, Map[BreakpointPileup, PileupGroup]) = {
+
+    // Identify the immediate neighbors of the given pileup, and start a cluster containing them
+    val neighbors: PileupGroup = pileupToNeighbors.getOrElse(pileup, Seq.empty)
+    val currCluster: PileupGroup = neighbors :+ pileup
```
nh13, line 146: where is `existingCluster` used?  I think it needs to be added here:
```suggestion
    val currCluster: PileupGroup = existingCluster ++ neighbors :+ pileup
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+
+    // Remove the given pileup and its immediate neighbors from `pileupToNeighbors`
```
nh13, line 148: I don't think the `neighbors` are removed yet
```suggestion
    // Remove the given pileup from `pileupToNeighbors`
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,281 @@
...
+          split_reads   = head.split_reads,
+          read_pairs    = head.read_pairs,
+          total         = head.total,
+        )
+        tail.foldLeft[AggregatedBreakpointPileup](headAgg)((agg, nextPileup) => agg.add(nextPileup))
+      case _ => throw new IllegalArgumentException("Pileup group must be non-empty")
+    }
+  }
+
+  private def addId(existing: String, add: String): String = f"${existing}_$add"
```
nh13, line 279: used only once, why not inline?

nh13 (changes requested):
Looking really good.  I think there's one bug I'd like you to look at.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
pamelarussell: We actually don't need `existingCluster` at all. It is a relic from the previous implementation. Thanks for catching.

pamelarussell (commented):
@nh13 the bug you saw was actually an unnecessary value that remained from the old implementation. Thanks for noticing. I added a set of breakpoints to the main end-to-end test that should have broken the test given what that bug would have been, but couldn't break it, because `existingClusters` wasn't being used. Keeping the expanded test.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+    val pileups: Seq[BreakpointPileup] = Metric.read[BreakpointPileup](input)
+    logger.info(f"Read ${pileups.length} pileups from file $input")
+
+    // Open bam reader
+    val samSource: Option[SamSource] = bam match {
+      case None => None
+      case Some(path) => Some(SamSource(path))
+    }
+
+    // Read targets from bed file and group by contig
```
tfenne, line 79: More generally I think you could load these up straight into an `OverlapDetector` (class from HTSJDK) which is basically a set of `IntervalTree`s, one per contig.  Then you can very efficiently query for overlaps.

tfenne: I think the OverlapDetector also works with anything that implements `Locatable` which includes `BEDFeature` so no need to convert to Interval either.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+        logger.info(f"Read ${contigToTargets.values.map(_.size).sum} targets on ${contigToTargets.keys.size} " +
+          f"contigs from file $bed")
+        Some(contigToTargets.toMap)
+    }
+
+    // Open output writer
+    val writer = Metric.writer[AggregatedBreakpointPileup](output)
+
+    // Group pileups by left and right contig and strand
+    pileups.groupMap(BreakpointContigOrientation(_))(identity).foreach { case (_, group) =>
```
tfenne, line 102: Why use `groupMap` with `identity` when `groupBy` is the same thing?  

```suggestion
    pileups.groupBy(BreakpointContigOrientation(_)).foreach { case (_, group) =>
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+        .map(_.calculateFrequency(samSource, flank, minTotal, minFrequency))
+        .map(_.addTargetOverlap(targets))
+        .foreach(writer.write)
+    }
+
+    writer.close()
+
+  }
+
+}
```
tfenne, line 125:
```suggestion
    writer.close()
  }
}
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+   * @param pileups Set of pileups to evaluate for pairwise neighbor relationships
+   * @param maxDist Maximum distance (inclusive) for the left and right coordinates to be considered a neighbor
+   * @return Map of pileup to the set of neighbors
+   */
+  private def pileupToNeighbors(pileups: PileupGroup, maxDist: Int): Map[BreakpointPileup, PileupGroup] = {
+    val map: mutable.Map[BreakpointPileup, List[BreakpointPileup]] = mutable.Map.from(pileups.map((_, Nil)))
+    for (pileup1 <- pileups; pileup2 <- pileups) {
+      if (pileup1.total >= pileup2.total && isCluster(pileup1, pileup2, maxDist)) {
+        map.put(pileup1, (pileup2 :: map(pileup1)).distinct)
+        map.put(pileup2, (pileup1 :: map(pileup2)).distinct)
```
tfenne, line 148: Why do you need distinct here?  It seems really expensive to do and I can't figure out why you need it at all?  If you really need it maybe just return:

```scala
map.iterator.map { case (k, vs) => (k, vs.distinct) }.toMap
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+   * to reduce unnecessary run time.
+   *
+   * @param pileups Set of pileups to evaluate for pairwise neighbor relationships
+   * @param maxDist Maximum distance (inclusive) for the left and right coordinates to be considered a neighbor
+   * @return Map of pileup to the set of neighbors
+   */
+  private def pileupToNeighbors(pileups: PileupGroup, maxDist: Int): Map[BreakpointPileup, PileupGroup] = {
+    val map: mutable.Map[BreakpointPileup, List[BreakpointPileup]] = mutable.Map.from(pileups.map((_, Nil)))
+    for (pileup1 <- pileups; pileup2 <- pileups) {
+      if (pileup1.total >= pileup2.total && isCluster(pileup1, pileup2, maxDist)) {
```
tfenne, line 146: Is it intentional to include `pileup1` in it's own pileup group?

tfenne: I still think it would be better to:
1. When you read in the metrics above do a `toIndexedSeq`
2. Redefine PileupGroup to `IndexedSeq[BreakpointPileup]`
3. Redo the loop below to something like:

```scala
for (i <- pileups.indices; j <- pileups.indices.drop(i) {
  val pileup1 = pileups(i)
  val pileup2 = pileups(j)
  ...
}
```

It significantly reduces the looping and eliminates the need for the `.distinct` calls I believe.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+    }
+    map.toMap
+  }
+
+  /** Returns true if the two breakpoints can be clustered, i.e., they are not the same breakpoint (have different
+   * IDs), they have the same left and right chromosomes and strand, and the distance between their left and right
+   * coordinates is below the threshold (inclusive) */
+  private def isCluster(bp1: BreakpointPileup, bp2: BreakpointPileup, maxDist: Int): Boolean = {
+    bp1.id != bp2.id &&
+      BreakpointContigOrientation(bp1) == BreakpointContigOrientation(bp2) &&
```
tfenne, line 159: There's an awful lot of creation of BreakpointContigOrientation objects here and elsewhere.  Perhaps this should just be a `val` or `lazy val` on BreakpointPileup` so we don't make it over and over?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+   *                     returned.
+   */
+  def calculateFrequency(source: Option[SamSource], flank: Int, minTotal: Int, minFrequency: Double):
+  AggregatedBreakpointPileup = {
+
+    def frequency(contig: String, pileups: PositionList, minPos: Int, maxPos: Int): Option[Double] = source match {
+      case None => None
+      case Some(src) =>
+        val numOverlap = numOverlappingTemplates(
+          src, contig, pileups, minPos - flank, maxPos + flank, minTotal, minFrequency)
```
tfenne, line 347: Would be nice to pass by name here given so many numeric parameters.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+   *                 execution when the number of supporting templates is too low to be interesting.
+   * @param minFrequency Minimum proportion of templates overlapping a breakend that support the breakpoint. Used to
+   *                     stop iteration over huge sets of overlapping templates when the number of templates supporting
+   *                     the breakpoint is insufficient to be interesting. When the number of overlapping templates
+   *                     causes (total_supporting_breakpoint / total_overlappers) to drop below this frequency, None is
+   *                     returned.
+   * @return The size of the union of templates that overlap any breakend
+   */
+  private def numOverlappingTemplates(source: SamSource, contig: String, breakends: PositionList, queryMin: Int,
+                                      queryMax: Int, minTotal: Int, minFrequency: Double): Option[Int] = {
```
tfenne, line 379:
```suggestion
  private def numOverlappingTemplates(source: SamSource,
                                      contig: String,
                                      breakends: PositionList,
                                      queryMin: Int,
                                      queryMax: Int,
                                      minTotal: Int,
                                      minFrequency: Double): Option[Int] = {
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -0,0 +1,495 @@
...
+   *                 frequency. Used to speed up execution when the number of supporting templates is too low to be
+   *                 interesting.
+   * @param minFrequency Minimum proportion of templates overlapping a breakend that support the breakpoint. Used to
+   *                     stop iteration over huge sets of overlapping templates when the number of templates supporting
+   *                     the breakpoint is insufficient to be interesting. When the number of overlapping templates
+   *                     causes (total_supporting_breakpoint / total_overlappers) to drop below this frequency, None is
+   *                     returned.
+   */
+  def calculateFrequency(source: Option[SamSource], flank: Int, minTotal: Int, minFrequency: Double):
+  AggregatedBreakpointPileup = {
```
tfenne, line 341:
```suggestion
  def calculateFrequency(source: Option[SamSource],
                         flank: Int,
                         minTotal: Int,
                         minFrequency: Double): AggregatedBreakpointPileup = {
```

PR #18, opened 2022-08-16
Fix Stack overflow for large clusters

nh13:
I'd rather have `extractClusterFor` be tail recursive or in a loop, but I couldn't figure out how to do that easily, so this will do.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
@@ -13,7 +13,6 @@ import htsjdk.samtools.util.{Interval, OverlapDetector}
 import htsjdk.tribble.bed.{BEDCodec, BEDFeature}
 import htsjdk.tribble.AbstractFeatureReader
 
-import scala.annotation.tailrec
 import scala.collection.mutable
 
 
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -182,14 +182,14 @@ object AggregateSvPileup {
...
    * @return (1) The complete cluster for the given pileup; (2) Reduced version of `pileupToNeighbors` with all pileups
    *         in the returned cluster having been removed from both keys and values.
    */
-  def extractClusterFor(pileup: BreakpointPileup, pileupToNeighbors: Map[BreakpointPileup, PileupGroup]):
+  def extractClusterFor(pileup: BreakpointPileup, pileupToNeighbors: Map[BreakpointPileup, PileupGroup], neighborsToIgnore: Set[BreakpointPileup]):
   (PileupGroup, Map[BreakpointPileup, PileupGroup]) = {
-
     // Identify the immediate neighbors of the given pileup, and start a cluster containing them
-    val neighbors: PileupGroup = pileupToNeighbors.getOrElse(pileup, IndexedSeq())
+    val neighbors: PileupGroup = pileupToNeighbors.getOrElse(pileup, IndexedSeq()).filter(neighborsToIgnore.contains)
```
nh13, line 192: this needs careful thought on correctness

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -182,14 +182,14 @@ object AggregateSvPileup {
    *
    * @param pileup The pileup representative of the cluster that will be returned
    * @param pileupToNeighbors Map of pileup to its set of neighbors
+   * @parma neighborsToIgnore Set of pileups to ignore, as they are being considered in parent calls to this method
```
pamelarussell, line 185:
```suggestion
   * @param neighborsToIgnore Set of pileups to ignore, as they are being considered in parent calls to this method
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
pamelarussell: @nh13 are you able to generate an example that causes a stack overflow and add it as a test case?

At a minimum, we should fix the compile errors in the [test file](https://github.com/fulcrumgenomics/fgsv/blob/main/src/test/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupTest.scala) and make sure we haven't broken any known cases.

pamelarussell (changes requested):

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -159,49 +158,34 @@ object AggregateSvPileup {
...
    */
-  @tailrec
-  def toClusters(pileupToNeighbors: Map[BreakpointPileup, PileupGroup],
-                 existingClusters: Seq[PileupGroup] = Seq()): Seq[PileupGroup] = {
-    pileupToNeighbors.headOption match {
-      case None              => existingClusters
-      case Some((pileup, _)) =>
-        val (pileupGroup, remaining) = extractClusterFor(pileup, pileupToNeighbors)
-        toClusters(remaining, existingClusters :+ pileupGroup)
+  def toClusters(pileupToNeighbors: Map[BreakpointPileup, PileupGroup]): Seq[PileupGroup] = {
```
nh13, line 163: Could not add a test case that made the previous implementation overflow due to restrictions on sharing that data.  Not sure it's worth it, but happy to add a few more test cases if you think that would be worth it.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -159,49 +158,34 @@ object AggregateSvPileup {
...
+    // The following finds connnected components in a breadth first manner
+    val visited    = scala.collection.mutable.HashSet[BreakpointPileup]()
+    val components = IndexedSeq.newBuilder[PileupGroup]
+    pileupToNeighbors.keys.foreach { rootPileup =>
+      if (!visited.contains(rootPileup)) {
+        val remaining = scala.collection.mutable.Queue[BreakpointPileup]()
+        val component = Set.newBuilder[BreakpointPileup]
+        remaining += rootPileup
```
nh13, line 171:
```suggestion
        remaining.enqueue(rootPileup)
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -159,49 +158,34 @@ object AggregateSvPileup {
...
+        while (remaining.nonEmpty) {
+          // get the next pileup to examine
+          val curPileup = remaining.dequeue()
+          if (!visited.contains(curPileup)) {
+            // add it to this component
+            component += curPileup
+            // add all unvisited neighbors
+            remaining ++= pileupToNeighbors(curPileup).filterNot(visited.contains)
```
nh13, line 179: I wonder if we can avoid enqueueing the same pileup multiple times if we mark it as visited after we've enqueued it, then remove if on line 175.
```suggestion
            val unvisited = pileupToNeighbors(curPileup).filterNot(visited.contains)
            remaining ++= unvisited
            unvisited.foreach(visited.add)
```

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -159,49 +158,36 @@ object AggregateSvPileup {
...
+          // get the next pileup to examine
+          val curPileup = remaining.dequeue()
+          // add it to this component
+          component += curPileup
+          // find all unvisited neighbors
+          val unvisited = pileupToNeighbors(curPileup).filterNot(visited.contains)
+          // add the unvisited neighbors to the queue
+          remaining ++= unvisited
+          // mark the unvisited neighbors as being visited so we don't enqueue them later
+          unvisited.foreach(visited.add)
```
nh13, line 182:
```suggestion
          visited ++= unvisited
```

src/test/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupTest.scala
```diff
@@ -231,22 +231,6 @@ class AggregateSvPileupTest extends UnitSpec {
     )
   }
 
-  "AggregateSvPileup.extractClusterFor" should "return correct cluster of size 1" in {
```
pamelarussell, line 234: Why remove these two tests?

pamelarussell (changes requested):

src/test/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupTest.scala
nh13: The method no longer exists.

pamelarussell (approved):

PR #20, opened 2023-05-11
Add option BED file to SvPileup

nh13:
The user may provide an optional BED file with `--targets-bed`.  The output breakpoints will be annotated with a comma-delimited list of targets in the BED file (by name) that overlap each side respectively.  Therefore, this changes the output format.

Furthermore, the `--targets-bed-requirement` option can be used to specify that only breakpoints that at least one target (at least one side via `OverlapsAny`, and both sides via `OverlapsBoth`) are written to the output.  In the case of `OverlapsBoth`, each side of a breakpoint does not have to overlap _the same_ target, just that they each independently overlap a target (can be a different target).

### DIFF
--- a/docs/04_Metrics.md
+++ b/docs/04_Metrics.md
@@ -66,8 +66,8 @@ the only information comes from read-pairs and the breakpoint information should
 |split_reads|Int|The number of templates/inserts with split-read alignments that identified this breakpoint.|
 |read_pairs|Int|The number of templates/inserts with read-pair alignments (and without split-read alignments)                      that identified this breakpoint.|
 |total|Int|The total number of templates/inserts that identified this breakpoint|
-|left_target|Option[String]|The comma-delimited list of target names overlapping the left breakpoint|
-|right_target|Option[String]|The comma-delimited list of target names overlapping the right breakpoint|
+|left_targets|Option[String]|The comma-delimited list of target names overlapping the left breakpoint|
+|right_targets|Option[String]|The comma-delimited list of target names overlapping the right breakpoint|
 
 
 ### MergedPileup
--- a/docs/04_Metrics.md
+++ b/docs/04_Metrics.md
@@ -66,6 +66,8 @@ the only information comes from read-pairs and the breakpoint information should
 |split_reads|Int|The number of templates/inserts with split-read alignments that identified this breakpoint.|
 |read_pairs|Int|The number of templates/inserts with read-pair alignments (and without split-read alignments)                      that identified this breakpoint.|
 |total|Int|The total number of templates/inserts that identified this breakpoint|
+|left_target|Option[String]|The comma-delimited list of target names overlapping the left breakpoint|
+|right_target|Option[String]|The comma-delimited list of target names overlapping the right breakpoint|
 
 
 ### MergedPileup
--- a/src/main/scala/com/fulcrumgenomics/sv/BreakpointPileup.scala
+++ b/src/main/scala/com/fulcrumgenomics/sv/BreakpointPileup.scala
@@ -22,6 +22,8 @@ import com.fulcrumgenomics.util.Metric
  * @param read_pairs    the number of templates/inserts with read-pair alignments (and without split-read alignments)
  *                      that identified this breakpoint.
  * @param total         the total number of templates/inserts that identified this breakpoint
+ * @param left_target   the comma-delimited list of target names overlapping the left breakpoint
+ * @param right_target  the comma-delimited list of target names overlapping the right breakpoint
  */
 case class BreakpointPileup(id: String,
                             left_contig: String,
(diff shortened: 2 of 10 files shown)


docs/04_Metrics.md
```diff
@@ -66,6 +66,8 @@ the only information comes from read-pairs and the breakpoint information should
 |split_reads|Int|The number of templates/inserts with split-read alignments that identified this breakpoint.|
 |read_pairs|Int|The number of templates/inserts with read-pair alignments (and without split-read alignments)                      that identified this breakpoint.|
 |total|Int|The total number of templates/inserts that identified this breakpoint|
+|left_target|Option[String]|The comma-delimited list of target names overlapping the left breakpoint|
```
tfenne, line 69: Should these be plural (`left_targets`)?

src/main/scala/com/fulcrumgenomics/sv/BreakpointPileup.scala
```diff
@@ -32,12 +34,15 @@ case class BreakpointPileup(id: String,
                             right_strand: Char,
                             split_reads: Int,
                             read_pairs: Int,
-                            total: Int
+                            total: Int,
+                            left_target: Option[String] = None,
```
tfenne, line 38: Same comment on target->targets

src/main/scala/com/fulcrumgenomics/sv/BreakpointPileup.scala
```diff
@@ -32,12 +34,15 @@ case class BreakpointPileup(id: String,
...
+                            right_target: Option[String] = None
                            ) extends Metric {
 
   override def toString(): String = f"$id|$left_contig:$left_pos($left_strand)/" +
     f"$right_contig:$right_pos($right_strand)|" +
-    f"$split_reads,$read_pairs,$total"
+    f"$split_reads,$read_pairs,$total" +
+    f"${left_target.getOrElse("")},${right_target.getOrElse("")}"
```
tfenne, line 45: CSVs within CSVs!
Seriously though - I think you a) need something between $total and ${left...}, and b) might want to to `left_target.map(_.replace(',' ';')` or similar so it's clear what's what.  Otherwise how do I tell what are the left and right targets?

src/main/scala/com/fulcrumgenomics/sv/FgSvDef.scala
```diff
@@ -0,0 +1,16 @@
+package com.fulcrumgenomics.sv
+
+import com.fulcrumgenomics.FgBioDef.FilePath
+import htsjdk.samtools.util.OverlapDetector
+import htsjdk.tribble.AbstractFeatureReader
+import htsjdk.tribble.bed.{BEDCodec, BEDFeature}
+
+object FgSvDef {
+  /** Builds an [[OverlapDetector]] frpom a BED file */
```
tfenne, line 9:
```suggestion
  /** Builds an [[OverlapDetector]] from a BED file */
```

src/main/scala/com/fulcrumgenomics/sv/FgSvDef.scala
```diff
@@ -0,0 +1,16 @@
...
+  def overlapDetectorFrom(bed: FilePath): OverlapDetector[BEDFeature] = {
```
tfenne, line 10: Is there a reason to have this in a `*Def`?  Why not just `Intervals.overlapDetectorFromBed()`?

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -127,19 +151,31 @@ class SvPileup
           slop                     = slop
         )
 
+        val filteredEvidences = evidences.filter { ev =>
+          targets.forall { detector =>
```
tfenne, line 155: I think I would put the `targets.forall` on the outside (and make it `targets.foreach`) so that you don't even do the `filter()` call if you don't have targets to filter to.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -172,13 +208,16 @@ class SvPileup
...
+                               dict: SequenceDictionary,
+                               targets: Option[OverlapDetector[BEDFeature]]): Unit = {
     val writer      = Metric.writer[BreakpointPileup](path)
     val breakpoints = tracker.breakpoints.toIndexedSeq.sortWith(Breakpoint.PairedOrdering.lt)
 
     breakpoints.foreach { bp =>
-      val info   = tracker(bp)
-      val metric = BreakpointPileup(
+      val leftTargets  = targets.map(_.getOverlaps(bp.leftInterval(dict)).map(_.getName).toSeq.sorted.mkString(","))
+      val rightTargets = targets.map(_.getOverlaps(bp.rightInterval(dict)).map(_.getName).toSeq.sorted.mkString(","))
```
tfenne, line 218: Might want to unique after the `.map(_.getName)` to be safe.

tfenne (approved):
Some minor feedback.  Go ahead and merge once you're happy.

PR #28, opened 2023-09-29
Consider circularity when finding intra-contig breakpoints

jdidion:
If the contig is circular, then we consider segment pairs where seg2.end < seg1.start (or seg1.end < seg2.start on the negative strand) to be normally oriented and calculate the distance between them relative to the origin. 

I think I also fixed another bug (for the non-circular case), changing `seg1.range.start < seg2.range.start` to `seg1.range.start < seg2.range.end` for the negative strand comparison.


jdidion:
Leaving this as draft for now because I don't have time to add a test case.

jdidion:
@tfenne tests added, ready for review

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -376,29 +381,48 @@ object SvPileup extends LazyLogging {
...
-    // any segment that "moves backwards" down the genome is odd, as genome position and read position should increase
-    // together.
-    if (seg1.positiveStrand != seg2.positiveStrand) true
-    else if (seg1.positiveStrand && seg2.range.start < seg1.range.end) true
-    else if (!seg1.positiveStrand && seg1.range.start < seg2.range.start) true
+    // the segments should all come out on the same strand. Therefore any difference in strand is odd.
+    val positive_strand = seg1.positiveStrand
+    if (positive_strand != seg2.positiveStrand) {
+      return true
+    }
```
tfenne, line 398: We try to avoid use of explicit `returns` in scala.  Could you reformat as something like:

```scala
positive_strand != seg2.positiveStrand || {
  // rest of function
}
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -376,29 +381,48 @@ object SvPileup extends LazyLogging {
...
+    // Otherwise, any segment that "moves backwards" down the genome is odd, as genome position and read position 
+    // should increase together (unless the contig is curcular).
```
tfenne, line 400:
```suggestion
    // should increase together (unless the contig is circular).
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -376,29 +381,48 @@ object SvPileup extends LazyLogging {
...
+    val contig = dict(seg1.range.refIndex)
+    val is_circular = contig.topology.contains(Topology.Circular)
+    if (!is_circular && (
+      (positive_strand && seg2.range.start < seg1.range.end) ||
+      (!positive_strand && seg1.range.start < seg2.range.end)
+    )) {
+      return true
+    }
+    // If the contig is curcular and the segments span the origin, treat them as contiguous when
```
tfenne, line 409:
```suggestion
    // If the contig is circular and the segments span the origin, treat them as contiguous when
```

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -376,29 +381,48 @@ object SvPileup extends LazyLogging {
...
+    
     // The way aligned segments are generated for a template, if we have all the reads in the expected orientation
-    // the segments should all come out on the same strand.  Therefore any difference in strand is odd.  In addition
-    // any segment that "moves backwards" down the genome is odd, as genome position and read position should increase
-    // together.
-    if (seg1.positiveStrand != seg2.positiveStrand) true
-    else if (seg1.positiveStrand && seg2.range.start < seg1.range.end) true
-    else if (!seg1.positiveStrand && seg1.range.start < seg2.range.start) true
+    // the segments should all come out on the same strand. Therefore any difference in strand is odd.
+    val positive_strand = seg1.positiveStrand
```
tfenne, line 395: I've spent so much time in python and rust recently I didn't even see if at first.
`positive_strand` -> `positiveStrand`

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -376,29 +381,48 @@ object SvPileup extends LazyLogging {
...
+    if (positive_strand != seg2.positiveStrand) {
+      return true
+    }
+    // Otherwise, any segment that "moves backwards" down the genome is odd, as genome position and read position 
+    // should increase together (unless the contig is curcular).
+    val contig = dict(seg1.range.refIndex)
+    val is_circular = contig.topology.contains(Topology.Circular)
```
tfenne, line 402: `is_circular` -> `isCircular`

tfenne (approved):
This looks great, thanks @jdidion!  Just a few minor comments on formatting.

src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
jdidion: same

tfenne (approved):

PR #34, opened 2024-03-13
Add a description of the toolkit to the README

clintval:
Enhancing the README to better communicate how this toolkit works and should be used.


README.md
```diff
@@ -7,15 +7,37 @@
...
-[github-badge]:        https://github.com/fulcrumgenomics/fgsv/actions/workflows/unittests.yaml/badge.svg
+[github-badge]:        https://github.com/fulcrumgenomics/fgsv/actions/workflows/unittests.yaml/badge.svg?branch=main
 [github-link]:         https://github.com/fulcrumgenomics/fgsv/actions/workflows/unittests.yaml
 [scala-badge]:         https://img.shields.io/badge/language-scala-c22d40.svg
 [scala-link]:          https://www.scala-lang.org/
 [license-badge]:       https://img.shields.io/badge/license-MIT-blue.svg
 [license-link]:        https://github.com/fulcrumgenomics/fgsv/blob/main/LICENSE
 
-Tools to find evidence for structural variation.
+Tools for calling breakpoints and exploring structural variation.
```
nh13, line 17: Can we find a better description than _calling_, as we are really just gathering evidence (like mpileup) and not assessing the evidence (with a PL/GL etc).  How about a middle ground:

> Tools to gather evidence for structural variation via breakpoint detection.

README.md
```diff
@@ -7,15 +7,37 @@
...
 
 ## Documentation
 
-Documentation can be found in the [docs folder](docs/01_Introduction.md)
+Documentation can be found in the [docs folder](docs/01_Introduction.md).
+
+## Introduction to the `fgsv` Toolkit
+
+The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a query-grouped BAM file as input and scans through each template one at a time, where a template is the full collection of reads and alignments from a single source molecule.
```
nh13, line 25:
```suggestion
The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a query-grouped BAM file as input and scans through each template one at a time, where a template is the full collection of reads and alignments from a single source molecule.
For example, a paired end read may have an alignment per end (e.g. for read one and read two).
```

nh13: I was trying to find a way to make this more relatable to the layman.

README.md
```diff
@@ -7,15 +7,37 @@
...
+Primary and supplemental alignments for a template are used to construct a “chain” of aligned sub-segments in a way that is order and strand-aware.
```
nh13, line 26: 1. supplemental -> supplementary (here and throughout)
2. consider a shout-out to the SAM spec or elsewhere for the definitions of primary/supplementary
```suggestion
Primary and supplementary alignments for a template are used to construct a “chain” of aligned sub-segments in a way that is aware of both order and strand.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+These aligned sub-segments relate to each other through typical alignment mechanisms like insertions and deletions but also contain information about the relative orientation of the sub-segment to the reference genome and importantly, jumps between reference sequences (chromosomes).
```
nh13, line 27: 1. This was confusing, since most folks think of indels _within_ an alignment (sub-segment), versus here where we're thinking _between_ sub-segments.  
2. Should we inversions, inter and intra contig translocations, not just the latter?

README.md
```diff
@@ -7,15 +7,37 @@
...
+
+For each chain of aligned sub-segments per template, outlier jumps are collected where the minimum inter-segment distance within a read must be 100bp (by default) or greater and the minimum inter-read distance per pair must be 1000bp (by default) or greater.
```
nh13, line 29:
```suggestion
For each chain of aligned sub-segments per template, outlier jumps are collected where the minimum inter-segment distance within a read must be 100bp (by default) or greater, and the minimum inter-read distance across reads (e.g. between reads in a paired end read) must be 1000bp (by default) or greater.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+In the case where there is both evidence for a split-read alignment and inter-read jump, the split-read alignment evidence is favored.
```
nh13, line 30:
```suggestion
In the case where there is both evidence for a split-read alignment and inter-read jump, the split-read alignment evidence is favored since it gives a precise breakpoint.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+At locations where these jumps occur, breakpoints are marked, and the breakpoints are given a unique ID based on the position of the breakpoint, the directionality of the left and right strands, and the other location the aligned sub-segment jumps to.
```
nh13, line 31: Besides the suggestion below, should we describe the "directionality of the left and right strands", since I am not sure what that means.  Do you mean the left and right _break ends_?  If so, then you also need to describe the difference between a breakpoint and breakend!

```suggestion
At locations where these jumps occur, breakpoints are collected and given a unique ID based on (1) their genomic position, (2) the directionality of the left and right strands, and (3) the other location to which the aligned sub-segment jumps.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+The output of this process is simply a pileup of candidate breakpoint locations.
```
nh13, line 32:
```suggestion
The output of this process is a collection of candidate breakpoint locations.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+The output of this tool is a metrics file tabulating the breakpoints and a BAM file with each alignment having custom tags that indicate which breakpoint the alignment supports (by breakpoint ID), if any.
```
nh13, line 33: the BAM file only contains reads supporting the breakpoint I think.  Please confirm, and if so, please update the description.

README.md
nh13: Thinking about this more, we need a breakpoint vs breakend description

README.md
```diff
@@ -7,15 +7,37 @@
...
+
+Because of variability in short-read sequence data and their alignments, evidence for a single breakpoint may span a few loci near the true breakpoint.
```
nh13, line 35: Perhaps expound upon this?  
1. If the breakpoint only has inter-read evidence, then the breakpoint can occur anywhere within the unobserved bases (for example)
2. Due to similarity (homology) around the breakpoint (and each breakend), it is not always possible to locate it precisely, especially when comparing it to a truth set (think left-alignment of short indel variants).

README.md
```diff
@@ -7,15 +7,37 @@
...
+The tool [`fgsv AggregateSvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/AggregateSvPileup.md) is used to coalesce nearby breakpoints into one call if they appear to belong to one breakpoint.
+This polishing step preserves true positive breakpoint calls and should reduce the number of false positive breakpoint calls.
```
nh13, line 37:
```suggestion
This polishing step preserves true positive breakpoint calls and intends to reduce the number of false positive breakpoint calls.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+Adjacent breakpoints are only merged if their left sides map to the same reference sequence, their right rides sides map to the same reference sequence, the strandedness of the left and right aligned sub-segments is the same, and their left and right positions are both within a given length threshold.
```
nh13, line 38:
```suggestion
Adjacent breakpoints are only merged if their left sides map to the same reference contig, their right rides sides map to the same reference contig, the strandedness of the left and right aligned sub-segments is the same, and their left and right genomic positions are both within a given length threshold.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+One shortcoming of the existing behavior, that should be corrected at some point, is that inter-read breakpoint evidence is considered similarly to inter-pair breakpoint evidence even though inter-read breakpoint evidence often has nucleotide-level alignment resolution and inter-pair breakpoint evidence does not.
```
nh13, line 39:
```suggestion
One shortcoming of the existing behavior, which should be corrected at some point, is that inter-read breakpoint evidence is considered similarly to inter-pair breakpoint evidence even though inter-read breakpoint evidence often has nucleotide-level alignment resolution and inter-pair breakpoint evidence does not.
```

README.md
```diff
@@ -7,15 +7,37 @@
...
+The output of this tool is a metrics file tabulating the coalesced breakpoints with all previous breakpoint IDs listed for the new breakpoint call and an estimation of the allele frequency of the call based on the alignments that support the breakpoint.
+
+The `fgsv` tools are an effective structural variant debugging toolkit but are not meant to be considered as a structural variant calling toolchain in-and-of-itself.
+Instead, it’s better to think of the `fgsv` toolkit as an effective  “breakpoint caller”.
```
nh13, line 43: Again, lets avoid "caller"
```suggestion
Instead, it’s better to think of the `fgsv` toolkit as an effective  “breakpoint caller”.
```

nh13 (changes requested):

clintval:
- [ ] Make it very clear coordinates are always 1-based inclusive in the output files unless the output file is a BEDPE file (https://github.com/fulcrumgenomics/fgsv/pull/40#issuecomment-2105147858)

clintval:
@nh13 ready for a re-review when you have time! Thanks!

README.md
```diff
@@ -17,8 +17,75 @@
...
 
-Documentation can be found in the [docs folder](docs/01_Introduction.md)
+Documentation can be found in the [docs folder](docs/01_Introduction.md).
+
+## Introduction to the `fgsv` Toolkit
+
+The `fgsv` toolkit contains tools for effective structural variant debugging but are not meant to be used as a structural variant calling toolchain in-and-of-itself.
+Instead, it is better to think of `fgsv` as an effective breakpoint detection and structural variant exploration toolkit.
+
+When describing structural variation, we use the term breakpoint to mean a junction between two loci and the term breakend to refer to one of the loci in a breakpoint.
```
msto, line 31:
```suggestion
When describing structural variation, we use the term *breakpoint* to mean a junction between two loci and the term *breakend* to refer to one of the loci in a breakpoint.
```

I like to italicize definitions

README.md
```diff
@@ -17,8 +17,75 @@
...
+Importantly, all point intervals (1-length) reported by this toolkit are 1-based inclusive from the perspective of the reference sequence.
+
+### `fgsv SvPileup`
+
+Collates a pileup of putative structural variant supporting reads.
```
msto, line 36:
```suggestion
Collates pileups of reads over putative SV breakpoints.
```

is this clearer? "putative structural variant supporting" is a mouthful to have between "pileup of" and "reads"

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+```console
+fgsv SvPileup \
+    --input sample.bam \
+    --output sample.svpileup
+```
+
+The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a query-grouped BAM file as input and scans through each template one at a time, where a template is the full collection of reads and alignments from a single source molecule.
```
msto, line 44:
```suggestion
The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a queryname-grouped BAM file as input and scans through each template one at a time, where a template is the full collection of reads and alignments from a single source molecule.
```

nit - I think "name-sorted" (even though the sort order isn't a stipulation) is a more familiar term, so would like to get "name" in there

msto, line 44:
```suggestion
The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a query-grouped BAM file as input and scans each group of alignments for structural variant evidence.
```

I think the awkwardness here indicates that we are attempting to load too many new definitions and concepts at once. I'd prefer to defer or jettison the use of the unfamiliar `template`

README.md
```diff
@@ -17,8 +17,75 @@
...
+For example, a paired-end read may have an alignment per read: one alignment for read 1 and another alignment for read 2.
+
+Primary and supplementary alignments for a template (see the [SAM Format Specification v1](https://samtools.github.io/hts-specs/SAMv1.pdf) for more information) are used to construct a “chain” of aligned sub-segments in a way that honors the logical ordering of sub-segments and their strandeness in relation to the reference sequence.
```
msto, line 47: introduce `template` here, or swap to `queryname/group`?

msto, line 47:
```suggestion
Primary and supplementary alignments for a template (see the [SAM Format Specification v1](https://samtools.github.io/hts-specs/SAMv1.pdf) for more information) are used to construct a “chain” of aligned sub-segments in a way that honors the logical ordering of sub-segments and their strandedness in relation to the reference sequence.
```

README.md
```diff
@@ -17,8 +17,75 @@
...
+These aligned sub-segments in a chain relate to each other through typical alignment mechanisms like insertions and deletions but also contain information about the relative orientation of the sub-segment to the reference sequence and importantly, jumps between reference sequences such as translocations between chromosomes or contigs.
+
+For each chain of aligned sub-segments per template, outlier jumps are collected where the minimum inter-segment distance within a read must be 100bp (by default) or greater, and the minimum inter-read distance across reads (e.g. between reads in a paired-end read) must be 1000bp (by default) or greater.
```
msto, line 50:
```suggestion
For each chain of aligned sub-segments, outlier jumps are collected where the minimum inter-segment distance within a read must be 100bp (by default) or greater, and the minimum inter-read distance across reads (e.g. between reads in a paired-end read) must be 1000bp (by default) or greater.
```

This is maybe not a question for this PR, but why were these thresholds chosen as opposed to either 50 bp (the traditional indel/SV threshold; for split reads) or the SAM proper pair flag (library insert size distribution dependent; for paired reads)

README.md
```diff
@@ -17,8 +17,75 @@
...
+In the case where there is both evidence for a split-read alignment and inter-read jump, the split-read alignment evidence is favored since it gives a precise breakpoint.
```
msto, line 51: What about the case where the breakpoints inferred by the split alignment and paired alignment are different? (edge case)

README.md
```diff
@@ -17,8 +17,75 @@
...
+At locations where these jumps occur, breakpoints are marked and the breakpoints are given a unique ID based on the positions of the breakends and the directionality of the left and right strands leading into each breakend.
+
+This process creates a collection of candidate breakpoint locations.
+The output of this tool is a metrics file tabulating the breakpoints and a BAM file with each breakpoint-supporting alignment having custom tags that indicate which breakpoint the alignment supports.
```
msto, line 55:
```suggestion
The tool outputs a table of breakpoints (in a BEDPE-like format) and a modified copy of the input BAM file where each alignment is tagged with the ID of the breakpoint it supports (if any).
```

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+### `fgsv AggregateSvPileup`
+
+Merges nearby pileups of reads supporting putative breakpoints.
```
msto, line 59: ?
```suggestion
Aggregates and merges pileups that are likely to support the same breakpoint. 
```

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+```console
+fgsv AggregateSvPileup \
+    --bam sample.bam \
+    --input sample.svpileup.txt \
+    --output sample.svpileup.aggregate.txt
+```
+
+Because of variability in typical short-read alignments, evidence for a single breakpoint may span a few loci near the true breakend loci. For example, if the breakpoint only has intra-read evidence, then the breakpoint could coincidentally occur within the unobserved bases between read 1 and read 2 in a pair. In other cases and due to sequence similarity or homology between each breakend locus, it is not always possible to locate the exact nucleotide point where the breakends occur, and instead a plausible region may exist that supports either breakend loci.
```
msto, line 68:
```suggestion
Because of variability in typical short-read alignments, evidence for a single breakpoint may span a few loci near the true breakend loci. For example, if the breakpoint only has paired-end evidence, then the breakpoint could coincidentally occur within the unobserved bases between read 1 and read 2 in a pair. In other cases and due to sequence similarity or homology between each breakend locus, it is not always possible to locate the exact nucleotide point where the breakends occur, and instead a plausible region may exist that supports either breakend loci.
```

or `inter-read`, but I find `split-read` and `paired-end` to be more familiar terms

msto: I've read this a few times and don't quite understand what it's saying

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+The tool [`fgsv AggregateSvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/AggregateSvPileup.md) is used to coalesce nearby breakpoints into one event if they appear to belong to one true breakpoint.
```
msto, line 70:
```suggestion
The tool [`fgsv AggregateSvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/AggregateSvPileup.md) is used to coalesce pileups in proximity to each other into one event if they appear to support the same breakpoint.
```

README.md
```diff
@@ -17,8 +17,75 @@
...
+This polishing step preserves true positive breakpoint events and intends to reduce the number of false positive breakpoint events.
```
msto, line 71:
```suggestion
This polishing step preserves true positive breakpoint events and is intended to reduce the number of false positive breakpoint events.
```

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+Adjacent breakpoints are only merged if their left breakends map to the same reference sequence, their right breakends map to the same reference sequence, the strandedness of the left and right aligned sub-segments is the same, and their left and right genomic breakend positions are both within a given length threshold.
```
msto, line 73:
```suggestion
Adjacent breakpoints are only merged if their left breakends map to the same strand of the same reference chromosome (or contig), their right breakends map to the same strand of the same reference chromosome (or contig), and their left and right genomic breakend positions are both within a given length threshold.
```

Also, what is the threshold?

README.md
```diff
@@ -17,8 +17,75 @@
...
+
+One shortcoming of the existing behavior, which should be corrected at some point, is that intra-read breakpoint evidence is considered similarly to inter-pair breakpoint evidence even though intra-read breakpoint evidence often has nucleotide-level alignment resolution and inter-pair breakpoint evidence does not.
+
+The output of this tool is a metrics file tabulating the coalesced breakpoints with all previous breakpoint IDs listed for the new breakpoint event and an estimation of the allele frequency of the event based on the alignments that support the breakpoint.
```
msto, line 77:
```suggestion
The output of this tool is a table of coalesced breakpoints. Each aggregate breakpoint is annotated with the constituent pileup IDs and an estimation of the allele frequency of the breakpoint based on the alignments that support the breakpoint.
```

How is allele frequency calculated? Especially without the original bam?

msto (changes requested):

README.md
clintval: I'm not sure. 1000 or ~1,200 are common numbers I've seen for improperly distanced short reads.

I went back through the Git history in this repo but found no documentation or reference for these values.

README.md
clintval: More detailed algorithmic details are documented in the tool-specific CLI documentation:

https://github.com/fulcrumgenomics/fgsv/blob/fbd4861da131e78aec3d64f96a92f34b7144720d/src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala#L93-L110

All other edge cases could be learned by reading the code itself.

My goal for this PR is to have a frontpage README that generally describes these tools.

More detailed documentation should exist at the tool-level. If edge case or granular behavior isn't documented, then we should avoid cluttering the main README and add that to the tool docs.

README.md
clintval: More detailed docs exist at the tool-level including a definition for allele frequency:

https://github.com/fulcrumgenomics/fgsv/blob/fbd4861da131e78aec3d64f96a92f34b7144720d/src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala#L37-L39

clintval:
Ready for another re-review with fresh eyes! 👀 

I had a few requests for a more detailed algorithmic overview than what is currently provided. My goal is to have the main README describe the gist of how the tools work and then we can add more detailed documentation in the _**tool-level**_ CLI help docs.

README.md
```diff
@@ -17,8 +17,82 @@
...
+Collates pileups of reads over breakpoint events.
+
+```console
+fgsv SvPileup \
+    --input sample.bam \
+    --output sample.svpileup
+```
+
+The tool [`fgsv SvPileup`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/SvPileup.md) takes a queryname-grouped BAM file as input and scans each query group (template) of alignments for structural variant evidence.
+For a simple example: a paired-end read may have one alignment per read—one alignment for read 1 and another alignment for read 2 mapped to different reference sequences supporting a putative translocation.
```
nh13, line 48: "may have one alignment per read—one alignment for read 1" doesn't parse for me, can you take a look?

README.md
```diff
@@ -17,8 +17,82 @@
...
+```console
+fgsv AggregateSvPileupToBedPE \
+    --input sample.svpileup.aggregate.txt \
+    --output sample.svpileup.aggregate.bedpe
+```
+
+The tool [`fgsv AggregateSvPileupToBedPE`](https://github.com/fulcrumgenomics/fgsv/blob/main/docs/tools/AggregateSvPileupToBedPE.md) is used to convert the output of `AggregateSvPileup` to BEDPE so that it can be viewed in [IGV](https://igv.org/) and other BEDPE-supporting genome browsers.
+For example:
+
+![BEDPE in IGV](docs/img/fgsv-bedpe.png)
```
nh13, line 98: pretty!

nh13 (approved):
Minor suggestions, take it or leave it

README.md
```diff
@@ -17,8 +17,82 @@
...
 
 ## Documentation
 
-Documentation can be found in the [docs folder](docs/01_Introduction.md)
+More detailed documentation can be found in the [docs folder](docs/01_Introduction.md).
+
+## Introduction to the `fgsv` Toolkit
+
+The `fgsv` toolkit contains tools for effective structural variant debugging.
+These tools are not meant to be used as a structural variant calling toolchain in-and-of-itself; instead, it is better to think of `fgsv` as an effective breakpoint detection and structural variant exploration toolkit.
```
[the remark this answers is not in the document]
clintval: I agree. It's not good form to be promotional in technical/scientific descriptions anyway.

README.md
clintval: Good catch.  Botched sentence due to a Git merge.

PR #37, opened 2024-04-10
Add a tool to make IGV-compat. BEDPE files from AggregateSvPileup

clintval:
Hopefully this addition is welcome! Converting the output of AggregateSvPileup to BEDPE facilitates genome-browsing in IGV and helps with understanding how breakpoint events relate to the alignments that cause them. There exist more advanced features of BEDPE in IGV (like record-specific coloring) but I chose not to implement that at this time to keep the tool simple and forwards compatible. Here is what test data looks like in IGV:

Screenshot 2024-04-10 at 12 38 26 PM


src/main/scala/com/fulcrumgenomics/sv/cmdline/ClpGroups.scala
```diff
@@ -5,10 +5,16 @@ import com.fulcrumgenomics.sopt.cmdline.ClpGroup
...
   }
 
-  final val All = classOf[_All]
+  class _Utilities  extends ClpGroup {
+    override val name: String = "Utility Tools"
+    override val description: String = "Helper tools for working with breakpoint or SV data."
+  }
+
+  final val BreakpointAndSv = classOf[_BreakpointAndSv]
+  final val Utilities = classOf[_Utilities]
```
clintval, line 19: This is a backwards compatible change and make the CLI look like:

Screenshot 2024-04-10 at 12 38 05 PM

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    /** Build a [[BedPE]] record from an [[AggregatedBreakpointPileup]]. */
+    def apply(pileup: AggregatedBreakpointPileup): BedPE = {
+      new BedPE(
+        chrom1  = pileup.left_contig,
+        start1  = pileup.left_min_pos,
+        end1    = pileup.left_max_pos + 1,
+        chrom2  = pileup.right_contig,
+        start2  = pileup.right_min_pos,
+        end2    = pileup.right_max_pos + 1,
```
clintval, line 75: Max positions are 0-based so to make then half-open, we have to add 1.

clintval, line 75: Max **_positions_** are 0-based so to make them half-open we need to add 1.

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  /** A simple BEDPE record as defined by `bedtools`:
+    *
+    * - https://bedtools.readthedocs.io/en/latest/content/general-usage.html#bedpe-format)
+    *
+    * Future compatibility could be implemented for supporting 10x flavored BEDPE files:
+    *
+    * - https://github.com/igvteam/igv/wiki/BedPE-Support
+    *
+    * Note that the field `score` is allowed to be a string per bedtools!
```
nh13, line 45: 'tis silly, and are we going to be ingesting strings?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
```diff
@@ -0,0 +1,127 @@
...
+        strand1 = Strand.decode(pileup.left_strand),
+        strand2 = Strand.decode(pileup.right_strand),
+      )
+    }
+
+    /** A writer class for writing [[BedPE]] records. */
+    class BedPEWriter(val out: BufferedWriter) extends Writer[BedPE] {
+
+      /** Write a [[BedPE]] record to the underlying writer. */
+      override def write(record: BedPE): Unit = {
```
nh13, line 94: this all feels a little clunky. Why not `out.write(record.values.mkString("\t"))` then write the newline?

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
```diff
@@ -0,0 +1,127 @@
...
+        start2  = pileup.right_min_pos,
+        end2    = pileup.right_max_pos + 1,
+        name    = pileup.id,
+        score   = pileup.total.toString,
+        strand1 = Strand.decode(pileup.left_strand),
+        strand2 = Strand.decode(pileup.right_strand),
+      )
+    }
+
+    /** A writer class for writing [[BedPE]] records. */
```
nh13, line 90: perhaps say that this is needed since we don't want the header

nh13 (approved):
nitpicks, but lgtm

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
clintval: forgot about that method!

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileupToBedPE.scala
clintval: Probably not, but leaving it as String means I don't have to pick a numeric type and right now I'm mapping total templates to the score (integers). I can switch it to Integer because no matter what we do in the future, if it is a change, it will be an API breaking change anyway.

PR #40, opened 2024-05-09
Fix coordinate conversion for BEDPE files

clintval:
I made a mistake and thought the coordinate system used here was 0-based.

It is 1-based! I've also added this clarification to the docs so others will benefit.


ameynert (approved):
Looks good to me!

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
```diff
@@ -230,12 +230,12 @@ object BreakpointCategory extends Enumeration {
...
- * @param left_min_pos          Minimum coordinate of left breakends
- * @param left_max_pos          Maximum coordinate of left breakends
+ * @param left_min_pos          Minimum coordinate of left breakends (1-based)
+ * @param left_max_pos          Maximum coordinate of left breakends (1-based)
  * @param left_strand           Strand at left breakends
  * @param right_contig          Contig name for right side of breakpoint
- * @param right_min_pos         Minimum coordinate of right breakends
- * @param right_max_pos         Maximum coordinate of right breakends
+ * @param right_min_pos         Minimum coordinate of right breakends (1-based)
+ * @param right_max_pos         Maximum coordinate of right breakends (1-based)
```
nh13, line 238: do we want to say inclusive here, or is that implied?

nh13 (commented):
qq about the coordinate system being 1-based inclusive.  Perhaps we should say that somewhere prominent (readme?)

src/main/scala/com/fulcrumgenomics/sv/tools/AggregateSvPileup.scala
clintval: Can't hurt, I'll add it!

clintval:
> qq about the coordinate system being 1-based inclusive. Perhaps we should say that somewhere prominent (readme?)

Good idea. I'll add this as as sub-task to https://github.com/fulcrumgenomics/fgsv/pull/34

PR #42, opened 2024-07-16
Defaulted to excluding duplicate and QC failing reads from pileup.

tfenne:


src/main/scala/com/fulcrumgenomics/sv/tools/SvPileup.scala
```diff
@@ -297,9 +305,11 @@ object SvPileup extends LazyLogging {
...
-    val r1PrimaryOk = t.r1.exists(r => r.mapped && r.mapq >= minPrimaryMapq)
-    val r2PrimaryOk = t.r2.exists(r => r.mapped && r.mapq >= minPrimaryMapq)
+                     minSupplementaryMapq: Int,
+                     incDupes: Boolean,
+                     incQcFails: Boolean): Option[Template] = {
+    val r1PrimaryOk = t.r1.exists(r => r.mapped && r.mapq >= minPrimaryMapq && (incDupes || !r.duplicate) && (incQcFails || r.pf))
+    val r2PrimaryOk = t.r2.exists(r => r.mapped && r.mapq >= minPrimaryMapq && (incDupes || !r.duplicate) && (incQcFails || r.pf))
 
     if (!r1PrimaryOk && !r2PrimaryOk) None else Some(
       Template(
```
msto, line 315: It's surprising to me that one read but not both may be filtered. I would expect the filtering would be applied to templates and not individual reads - should this be updated to filter the entire template if one read fails the mapq check? (Or to keep the entire template - perhaps the desired behavior could be a CLI flag?)

tfenne: I think that's probably a topic for a follow-up PR if we prefer.  That said, I think it does make sense that _reads_ can be filtered independently based on mapq at least.

E.g. I think you'd want to retain reads in the following situations:

1. Read 1 is mapped at high mapq and read 2 is unmapped
2. Read 1 is mapped at high mapq and read 2 has low mapq because it's wholly contained within a simple (dinuc/trinuc) repeat

In both cases, if R1 has a split read alignment then that information is still useful, no?

nh13 (approved):
lgtm

ameynert:
I ran the new version against the [example data](https://fulcrumgenomics.slack.com/archives/C0396BD6EMQ/p1721068312353899?thread_ts=1721061911.707939&cid=C0396BD6EMQ).

Lines output:

- `samtools markdup` marking duplicates only, prior to this change: 2030
- `samtools markdup` marking duplicates only, after to this change: 588
- `samtools markdup | samtools view -F 0x400`, both prior to and after this change: 550

When I took a look at what the differences were, I noticed the read support reported appears to include the duplicate reads, e.g. (top line soft marked, bottom line hard filtered)

```
516c481
< chr3	46373011	-	chr19	5452220	+	126	3	129
---
> chr3	46373011	-	chr19	5452220	+	34	3	37
```

If the user wants to exclude duplicates, then only the unmarked reads should be counted.

ameynert (changes requested):
Read support reporting shouldn't count the duplicate-marked reads.

msto:
I'm also surprised that the number of pileups changes when duplicate reads are excluded - given that pileups are reported with any number of supporting reads, I would expect the number of pileups to be constant while the number of supporting reads per pileup may decrease

docs/tools/index.md
```diff
@@ -4,7 +4,7 @@ title: fgsv tools
 
 # fgsv tools
 
-The following tools are available in fgsv version 0.2.0-d603e95.
+The following tools are available in fgsv version 0.2.0-5ce8bc6.
```
msto, line 7: Could you please bump the major version, as excluding duplicates (and qc fails) by default willl be a breaking change?

msto (changes requested):

PR #50, opened 2025-12-18
docs: replace tools symlink with direct references

nh13:


znorgaard (approved):
New links work for me!

PR #53, opened 2025-12-18
build: upgrade fgbio to 3.1.1

nh13:
Changes:                                                                                                                                                                                    
  1. build.sc: fgbio 2.2.1 → 3.1.1                                                                                                                                                            
  2. FgSvMetricsDoclet.scala: Added reporter parameter to constructor for fgbio 3.x compatibility                                                                                             
  3. AggregateSvPileupToBedPETest.scala: Fixed test to use explicit field names instead of reflection (fgbio 3.x's Metric trait now has internal fields that were being picked up

### DIFF
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -4,7 +4,7 @@ title: fgsv tools
 
 # fgsv tools
 
-The following tools are available in fgsv version 0.2.1-9885e88.
+The following tools are available in fgsv version 0.2.1-92b1669.
 ## Breakpoint and SV Tools
 
 Primary tools for calling and transforming breakpoints and SVs.


znorgaard (approved):
Works for me. I'm assuming we're not worried about the `BedPE` class getting out of sync with the `BedPEFields` defined for the test because it's a specification defined elsewhere.

clintval:
> Works for me. I'm assuming we're not worried about the `BedPE` class getting out of sync with the `BedPEFields` defined for the test because it's a specification defined elsewhere.

I think it's OK since it's defined only in unit tests. If something breaks in the future, it'll only break in tests and when users are running.

clintval (approved):

PR #63, opened 2026-02-20
fix: sort aggregated pileup IDs numerically instead of lexicographically

nh13:
## Summary
- Compound breakpoint IDs in `AggregatedBreakpointPileup` were sorted as strings, producing non-intuitive ordering (e.g. `"112_456_5_7_8_9"` instead of `"5_7_8_9_112_456"`)
- Change `pileupIds()` and `add()` to use `sortBy(_.toInt)` for numeric ordering
- Update test expectations to match new numeric ordering

## Test plan
- [ ] Existing tests pass with updated expectations (`./mill tools.test`)

### DIFF
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -4,7 +4,7 @@ title: fgsv tools
 
 # fgsv tools
 
-The following tools are available in fgsv version 0.2.1-b54e45d.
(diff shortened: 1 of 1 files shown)


clintval:
May also help with determinism. Thanks!

